PR #14, opened 2017-04-06
create draft of sdk documentation

smola:


driver/sdk.md
```diff
@@ -1,4 +1,97 @@
...
+```bash
+$ cd $GOPATH/src/github.com/bblfsh
+$ git init mylang-driver
+$ cd mylang-driver
+$ touch README.md
+$ git add README.md
+$ git commit -m 'initialize repository'
+```
+
+Now the driver should be bootstrapped with `bblfsh-sdk`:
```
alcortesm: This needs an explanation about why we are asked to do this, and what it does.  Something along these lines will be ok:
```
This will create some files and directories needed by future invocations of bblfsh-sdk,
overwriting them if they already exist, like the README.md file in the example below.
```

driver/sdk.md
```diff
@@ -1,4 +1,97 @@
...
+
+* [Docker](https://www.docker.com/get-docker)
+* [Go](https://golang.org/dl/)
+
+## Installing
+
+Install the Babelfish SDK:
+
+```bash
+$ go get github.com/bblfsh/sdk/...
```
alcortesm: Shouldn't we be using gopkg.in with versioning instead of github?

driver/kickoff.md
```diff
@@ -1,6 +0,0 @@
-# Creating a Driver
```
mcuadros: why we are removing this? I really thing that first we should explain, is the architecture and the constraints without using examples. An example is a indirect way to document something and once you already know how to use it, is hard to read to verify something

driver/sdk.md
alcortesm: Also, the insertion of the git hook caught me by surprise and made me lost some time when checking out a commit prior to running bblfsh-sdk, as I could no longer commit due to missing requirements of the hook. We should explain the this here.

driver/sdk.md
smola: Well, that's a good point. We should discuss if we want to version the SDK. Anyway, this documentation reflects the current state of the project.

driver/sdk.md
smola: :+1:

driver/kickoff.md
smola: :+1:

smola: Isn't that what's already explained in the architecture section? See https://doc.bblf.sh/architecture.html

driver/sdk.md
```diff
@@ -1,4 +1,135 @@
...
+```
+
+This will fetch the sdk repository to `$GOPATH/src/github.com/bblfsh/sdk` and
+will install the `bblfsh-sdk` CLI to `$GOPATH/bin/bblfsh-sdk`.
+
+## Bootstrapping a new driver
```
bzz: May be "Create a driver for a new language" would be a bit less ambiguous as bootstrap? 

Otherwise it may have a co-notation on "bootstrapping a compiler" and may refer to self-hosting, which seems not to be the case here.

driver/sdk.md
```diff
@@ -1,4 +1,135 @@
...
+creating file ".travis.yml"
+creating file "Dockerfile.build.tpl"
+creating file "driver/normalizer/normalizer_test.go"
+creating file "Dockerfile.tpl"
+creating file "LICENSE"
+managed file "README.md" has changed, discarding changes
+```
+
+Note that this adds a pre-commit [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks),
+which will verify these files are up to date before every commit. You can
```
bzz: > which will verify these files are up to date

may be it is worth mentioning up to date with what will these files be? Like where is the source of truth.

Also, have not seen `bblfsh-sdk update` mentioned/explained anywhere, so it might be a good spot to do so.

Something like:
```
Note that this adds a pre-commit [Git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks),
which will verify that these files are up to date with the [SDK /etc/skeleton templates](https://github.com/bblfsh/sdk/tree/master/etc/skeleton) before every commit.
You can update them manually anytime with `bblfsh-sdk update` or bypass this check with `git commit --no-verify`.
```

driver/sdk.md
smola: :+1:

SUMMARY.md
```diff
@@ -15,7 +15,6 @@
 
 ## Writing a Driver
 
-* [Creating a Driver](driver/kickoff.md)
```
bzz: May be for the visibility we could keep this section here, but point it to our new [./driver/sdk.md@bootstrapping-a-new-driver](https://github.com/bblfsh/documentation/pull/14/files#diff-7df754cc1aa08d4f35abb709f2592434R25) section?

driver/sdk.md
smola: :+1: added a new section for update

SUMMARY.md
smola: It seems weird to have two adjacent links pointing to the same page...

smola:
I'm merging this for now, and will prepare another PR about driver architecture, which should address @mcuadros' concerns.

PR #15, opened 2017-04-18
Initial versions of user quickstart docs

bzz:
For a user to quickly have a taste of what Babelfish is, it would be nice to have short tl;dr for getting from Source -> UAST

This is a temporary attempt to do so, for the current status of the project. It is expected to be updated, covering the same case, as project matures, docker images got published to dockerHub, etc.

The Run section right now does not work as described, but if people see value in it, I would be happy to submit a PR to [bblfsh/sdk/protocol/driver.go#L57](https://github.com/bblfsh/sdk/blob/master/protocol/driver.go#L57) adding something like this `parse-native-stdin`.

Please, let me know what you think!

### DIFF
--- a/user/getting-started.md
+++ b/user/getting-started.md
@@ -18,6 +18,7 @@ go get -u github.com/bblfsh/sdk/...
 
 # build a driver + container
 git clone https://github.com/bblfsh/java-driver.git
+go get -v -t ./...
 make build
 ```
 
--- a/user/getting-started.md
+++ b/user/getting-started.md
@@ -1,4 +1,35 @@
 
 # Getting Started
 
-* **TODO**
+Right now, 2017Q2 the easiest way to get started is to use local driver development environment: manually clone and build a container for a language driver with an SDK, then run it through Docker in order to get source -> UAST.
+
+I.e for a [Java driver](https://github.com/bblfsh/java-driver) :
+
+## Pre-requests
+ - Docker
+ - Go, for SKD build
+ - (macOS) `coreutils` and `gettext`, for SDK to work
+
+
+## Build
+```
+# build SDK
+go get -u github.com/bblfsh/sdk/...
+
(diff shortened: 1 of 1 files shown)


user/getting-started.md
```diff
@@ -1,4 +1,35 @@
...
+make build
+```
+
+In future, this will be replaced by `docker fetch` for the pre-build language driver containers
+
+## Run
+Get latest commit hash and run
+
+```
+cat 'native/src/main/java/bblfsh/Request.java' | docker run -it bblfsh/java-driver:dev-<commit[:6]> parse-native
```
smola: Feel free to send a PR so that if parse-native has an option (e.g. `--stdin`) it reads stdin instead of a file.

user/getting-started.md
```diff
@@ -1,4 +1,35 @@
...
-* **TODO**
\ No newline at end of file
+Right now, 2017Q2 the easiest way to get started is to use local driver development environment: manually clone and build a container for a language driver with an SDK, then run it through Docker in order to get source -> UAST.
+
+I.e for a [Java driver](https://github.com/bblfsh/java-driver) :
+
+## Pre-requests
+ - Docker
+ - Go, for SKD build
+ - (macOS) `coreutils` and `gettext`, for SDK to work
```
smola: s/macOS/OS X/ or s/macOS/Mac OS X/

bzz: I believe OS used to be called "OS X", but since last release it's official name going further is [macOS](https://en.wikipedia.org/wiki/Macintosh_operating_systems#Mac_OS_X.2FOS_X.2FmacOS) and that's why it was used. 

Should use the old one 👎  or a new one 👍  ?

smola: New one, since `macOS` seems to be what the cool guys use now :wink:

PR #21, opened 2017-05-26
[BIP-0] Index, template and the first BIP draft

abeaumont:
An initial draft for Babelfish Improvement Proposals.

### DIFF
--- a/proposals/bip-001.md
+++ b/proposals/bip-001.md
@@ -117,5 +117,5 @@ Each BIP must begin with a header that includes the following information:
 ## References
 
 This document is inspired by similar documents in other projects, specially
-[DEP](https://github.com/dylan-lang/website/blob/master/source/proposals/index.rst)
-and [PEP](https://www.python.org/dev/peps/).
+[DEP](https://opendylan.org/proposals/) and
+[PEP](https://www.python.org/dev/peps/).
--- a/proposals/bip-001.md
+++ b/proposals/bip-001.md
@@ -12,10 +12,19 @@
 ## Abstract
 
 BIP stands for Babelfish Improvement Proposal. A BIP is a design document which
-describes a change proposal to Babelfish. It's initially designed specifically
-to propose changes to [UAST](https://doc.bblf.sh/uast/specification.html)
-components regarding code categorization through Roles or related entities, but
-its scope may become broader in the future. It's informally defined.
+describes a change proposal to Babelfish. It's informally defined. The scope of
+a BIP includes:
+
+- Changes to [UAST](https://doc.bblf.sh/uast/specification.html)
+  components regarding code categorization through Roles or related entities.
+
+- Changes to Babelfish protocols (i. e. the
+  [server protocol](https://doc.bblf.sh/user/server-protocol.html),
+  [driver protocol](https://doc.bblf.sh/driver/protocol.html),
+  [internal driver protocol](https://doc.bblf.sh/driver/internal-protocol.html),
+  etc.
+
+Its scope may become broader in the future.
 
 The BIP should provide a concise technical specification of, and rational for,
 the proposed change.
@@ -33,9 +42,8 @@ need is not the best background for a universal feature request discussion, and
 it easies a process that may degenerate in a UAST that becomes the sum of each
 single language feature, which is something we should avoid. That's why
 introducing a more formal process, requiring a larger background and use cases,
-should help visualising which features are general enough, which ones are just
-special cases of a more general ones, which ones are just details that may be
-left for a future version, etc.
+should help visualising which features are general enough, how to generalize
+them and how to handle those features that are more specific.
 
 ## Workflow
 
@@ -47,11 +55,15 @@ The author must:
   is not guaranteed to be rejected.
 - Present a draft. It should take the next available BIP number and its status
   should be **Draft**
-- Reach consensus. Feedback should be added back to the proposal.
+- Reach consensus. Feedback should be added back to the proposal. You should
+  open a Pull Request with the proposal in the
+  [babelfish documentation project](https://github.com/bblfsh/documentation/)
+  for this process.
 
 Once the BIP has been completed and reviewed it may be **Accepted** or
 **Rejected** (it may have not been a good idea after all). In any case it's
 important to have a record of the discussion so the proposal will be archived.
+The final decision will be done by [mcuadros](https://github.com/mcuadros).
 
 
 ## Specification
@@ -99,6 +111,8 @@ Each BIP must begin with a header that includes the following information:
 - Status. One of Draft | Accepted | Rejected
 - Created. Date of creation
 - Updated. Date of last update
+- Target version. Target version for this feature to be implemented. It may not
+  appear if the BIP is not about a releaseable feature.
 
 ## References
 
--- a/proposals/bip-001.md
+++ b/proposals/bip-001.md
@@ -1,0 +1,107 @@
+# BIP Purpose and Guidelines
+
+| Field | Value |
+| --- | --- |
+| BIP | 1 |
+| Title | BIP Purpose and Guidelines |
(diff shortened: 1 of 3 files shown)


proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+| Created | 2017-05-26 |
+| Updated | 2017-05-26 |
+
+## Abstract
+
+BIP stands for Babelfish Improvement Proposal. A BIP is a design document which
+describes a change proposal to Babelfish. It's initially designed specifically
+to propose changes to [UAST](https://doc.bblf.sh/uast/specification.html)
+components regarding code categorization through Roles or related entities, but
+its scope may become broader in the future. It's informally defined.
```
abeaumont: I don't think we should use BIPs for every change in Babelfish, only for those which require some deeper background and discussion. Features in utils/ are just a show case and their changes shouldn't be formalised. API changes that break backward compatibility may be included in BIPs, but I'd wait till the API is stabilized before formalising its changes.

proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+to existing ones in the UAST components.
+
+## Rationale
+
+Currently changes to UAST are mostly language driven, i. e., whenever a language
+driver implementor finds that there's some feature missing to complete an AST to
+UAST conversion, this new feature is requested to be added through a Pull
+Request. While this process is quick, it doesn't scale well: a single language
+need is not the best background for a universal feature request discussion, and
+it easies a process that may degenerate in a UAST that becomes the sum of each
```
abeaumont: I don't see the improvement with the rewording, can you elaborate?

proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+single language feature, which is something we should avoid. That's why
+introducing a more formal process, requiring a larger background and use cases,
+should help visualising which features are general enough, which ones are just
+special cases of a more general ones, which ones are just details that may be
+left for a future version, etc.
```
abeaumont: I think that every feature shold be mapped to a general category, they should be as many categories as needed to achieve that, but not more. What we should not try to achieve is to have any single feature in every language as its own category, since that's unmanageable. That's only my opinion, it'd be nice to know if we all agreen on this, and then I can add some comments clearifying this point.

proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Workflow
+
+A BIP process should begin with an author proposing an idea. This idea should be
+as focused as possible.
+
+The author must:
+- Discuss the idea publicly, gather early feedback and make sure that the idea
```
abeaumont: I think the first step should be to discuss the idea _before_ a BIP process is started, to receive feedback and check if it makes sense. I don't think we want an archive with dozens of refjected BIPs which don't make sense at all.

proposals/bip-001.md
juanjux: I think that you're right about the utils but changed in the driver creation workflow, SDK public API (even if they don't break backwards compatibility, they could polute the API) should be included (it's Babelfish IP after all). We could probably add a "Scope" section indicating what parts of babelfish the BIP touches.

proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+  is not guaranteed to be rejected.
+- Present a draft. It should take the next available BIP number and its status
+  should be **Draft**
+- Reach consensus. Feedback should be added back to the proposal.
```
abeaumont: This is left undefined on purpose, since I don't think we need to formalize it. I think that a normal PR process can work well enough, but we can see how it works. If we see the need to formalize it, we can do it in the future. If we think we should do it now, I can change it aswell.

proposals/bip-001.md
juanjux: Not general categories, but if we create a new format with sublevels the language specific categories could be napped there. What is not an option, or that's the way I understood it, is to lose semantic information trying.

proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+
+Once the BIP has been completed and reviewed it may be **Accepted** or
+**Rejected** (it may have not been a good idea after all). In any case it's
```
abeaumont: Left undefined on purpose, I don't think we should formalise the process (at this point). At the end, the maintainers of the project (with merge permission) will obviously have the final word, but ideally it'd have been already decided by consensus in the review phase.

proposals/bip-001.md
juanjux: Oh ok I tough the discuss part was after making a PR for a BIP.

proposals/bip-001.md
```diff
@@ -0,0 +1,107 @@
...
+
+Each BIP must begin with a header that includes the following information:
+
+- BIP number
+- Title
+- Author. List of authors with real names.
+- Status. One of Draft | Accepted | Rejected
+- Created. Date of creation
+- Updated. Date of last update
+
```
abeaumont: The proposals should be timeless IMHO. A BIP may include information about what version of the UAST includes the implementation though.

proposals/bip-001.md
juanjux: Nah, I understood the sentence in another way, it's fine.

proposals/bip-001.md
juanjux: I think clarifying from the start how it's accepted or rejected could avoid conflicts later with people in the community making PRs and having them rejected, but it could be the matter of another BIP.

proposals/bip-001.md
juanjux: Could be left to another BIP like the comment below, but it should be clarified eventually.

proposals/bip-001.md
abeaumont: Ok, we can extend the scope. I don't think we need a scope section, it should be clear from the title or abstract.

proposals/bip-001.md
abeaumont: As I see it, consensus should be reach by any necessary means, that may be PR, slack, a mailing list or whatever other medium available in the future. I don't see much gain to state this explicitely, but I can do it.

proposals/bip-001.md
abeaumont: If we're going to specify it, I think this is the proper BIP. In such a case, we should define which will be the governing model of babelfish first, I guess @mcuadros and/or @eiso should be involved in that decision. As a reference, in the case of DEPs, this is left undefined, in the case of PEPs, the decision is made by BDFL (aka GvR).

eiso: The final decision on this is @mcuadros. My opinion is that in beginning we should have a "benevolent dictator" and as the project scales reevaluate the governing options.

proposals/bip-001.md
smola: I think we'll use the PR for initial discussions, but we'll probably have to refine the process.

proposals/bip-001.md
smola: :+1: the ultimate approval is approval by project admins, that is @mcuadros or whoever he delegates to.
However, everyone should voice his opinion on approval/rejection during the discussion. Most of the time, the admin should just follow the consensus if there is any.

proposals/bip-001.md
smola: :+1: to include tracking information of versions where it takes effect or where it is deprecated.

proposals/bip-001.md
mcuadros: "benevolent dictator"  makes sense

proposals/bip-001.md
smola: I agree that including full API of every component is just too much, but what about Babelfish protocols (server and driver protocols). BIPs should be more about specs rather than implementation. So for me, it would not make sense to include all Go APIs, but it would make sense to include full protocol specs. Basically, it's everything that ends up in a proto file.

PR #26, opened 2017-06-01
update getting started using babelfish server

smola:


user/protocol.md
```diff
@@ -0,0 +1,5 @@
+
+# Babelfish Protocol
+
+The Babelfish Protocol uses [grpc](http://www.grpc.io/). Check out the
```
smola: :+1:

user/getting-started.md
```diff
@@ -1,51 +1,84 @@
...
-make build
+Now you can test that it works by submitting a file for parsing:
+ 
 ```
+$ echo "import foo" > sample.py
+$ docker run -v $(pwd):/work --link bblfsh bblfsh/server bblfsh client --address=bblfsh:9432 /work/sample.py
+```
+
+First request might timeout, since the server has to fetch the required driver
+before responding. If it does, just retry.
```
abeaumont: During the process the server will log what it's doing. It may be useful to note this here, since it makes it easier to check that the delay is due to getting drivers.

user/getting-started.md
```diff
@@ -1,51 +1,84 @@
...
-# Get the hash of the docker container you just built
-docker images | grep python-driver
 
-# Replace the hash (the first 7 chars of the latest commit in the driver project)
-docker run -it --rm -v $GOPATH/src/github.com/bblfsh/python-driver:/code bblfsh/python-driver:dev-<commit[:7]>/opt/driver/bin/driver parse-uast --format=prettyjson /code/tests/sources/comprehension.py
+This will install the `bblfsh` binary into `$GOPATH/bin`.
 
-# Example with hash included that gets the UAST for the given `.py` file:
-docker run -it --rm -v $GOPATH/src/github.com/bblfsh/python-driver:/code bblfsh/python-driver:dev-955fc5d /opt/driver/bin/driver parse-uast --format=prettyjson /code/tests/sources/comprehension.py
+**[Linux only!]** Run the server:
```
abeaumont: Why linux only? If it'll only work with linux it'd be nice to indicate it at the beginning of the section, and explain why is that and if it may (or not) change in the future.

smola: The server requires Linux. That is why Windows and macOS users need to use a virtual machine. The recommended way to run it on Windows and macOS is using Docker for Windows and Docker for Mac (which handle the VM under the hood).

On why: the server uses containers to run plugins. Even if we add support for a container-like mode that does not really use Linux namespaces and cgroups, driver images would still be based on Linux images (e.g. Alpine Linux). I foresee that, in the future, we might be able to run the server on FreeBSD and Windows+WSL (*Windows Subsystem for Linux*) or use an embedded and lightweight VM.

user/getting-started.md
smola: :+1:

PR #28, opened 2017-06-01
[BIP-2] Explicit Annotated-Status and completeness information

juanjux:
Draft open for critic and discussion.

### DIFF
--- a/proposals/bip-002.md
+++ b/proposals/bip-002.md
@@ -100,7 +100,9 @@ tests.
 - The integration-test implementation would have to be modified to provide the
   summary of annotation completeness at the end of the run. This basis for this
   report would be implemented as an SDK tool that then would be called at the
-  end of the integration tests.
+  end of the integration tests. It's important to note that this report would
+  show coverage in the sense of unnanotated or incomplete nodes, but the annotated
+  nodes could still be wrongly annotated.
 
 
 ## References


proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+| Created | 2017-06-01 |
+| Updated | 2017-06-01 |
+
+## Abstract
+
+This BIP proposed a new official (as in defined in the Node struct and
+documented) "AnnotationStatus" property that would mark if the role has been
+annotated in any way and the quality of the current annotation.
+
+## Rationale
```
abeaumont: I miss some explanation here of the added complexity and how its benefits outperm it.

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+providing incomplete and even misleading information to UAST users that could
+treat coroutines differently (for example when making stack call deep analysis).
+
+## Specification
+
+This would modify the current specification by adding a new `AnnotationStatus`
+property into the `Node` struct (and the UAST nodes) and a `const` enum with
+the possible values.
+
+## Alternatives
```
abeaumont: I think of at least 2 more alternatives:
- Using just an `incomplete` property
- Not using annotation status at all

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+
+Instead of a new property, "Unannotated" and "Incomplete" could be UAST "meta"
```
abeaumont: How is this alternative worse than the proposed one?

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+much ground as possible. 
+
+This is a worthy objective but has the problem of leaving the nodes related with
+these features unannotated or with annotations that don't fully express all
+the semantic information that the language source code or the native AST have.
+
+This BIP proposes to add a new property in the Nodes called "AnnotationStatus"
+with the possible values of "Annotated", "Unannotated" and "Incomplete". This will
+make the UAST user have a wider view of the completeness of the information
+provided by it on specific nodes and also could help debug drivers and provide a
```
juanjux: Because nodes that have been incorrectly gone unnanotated (the most common bug when you're developing the annotations for a driver initially) will show even on the first generation of the integration-tests.

proposals/bip-002.md
juanjux: I think both are explained in along the BIP but I would add a micro-summary here.

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+completeness summary, with the added benefit of providing a numerically
+constatable measure of the completeness of every developed driver.
```
juanjux: Precisely, if the UAST definition change a node name but the driver developer don't update the driver it won't be annotated correctly and will show as such.

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+The Unannotated value for that property could be arguably less needed because it
```
juanjux: Yes, that's an alternative, but annotating as unannotated by default is basically free, and completes the AnnotationStatus property (which value do we add to it for unnanotated nodes if we remove unnanotated status?).

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+
+If a version of the UAST specification didn't have a role that could be matched
+for a "coroutine" type function supported by some language, the UAST could still
+mark those nodes as "FunctionDef", but without additional information, this would be
+providing incomplete and even misleading information to UAST users that could
+treat coroutines differently (for example when making stack call deep analysis).
+
+## Specification
+
+This would modify the current specification by adding a new `AnnotationStatus`
```
juanjux: - Status = Unnanotated: set by the SDK visitor by default when visiting nodes (driver programmer doesn't have to do anything).
- Status = Annotated: set by default by the SDK visitor as soon as any role is added. (ditto).
- Status = Incomplete: set by the driver programmer when adding other roles that he feels that doesn't correctly express the full semantic information.

Only the last one must be maintained, in the sense that if an UAST role is added that is a good fit for a previously Incomplete node the driver programmer must remove the setting of that status.

proposals/bip-002.md
juanjux: That was my first design, but IMHO it adds roles of a new category that doesn't provide semantic information and since we're already using properties for some other meta information (like position or original AST roles) I think properties are the natural fit for this status (that is also metainformation).

Also, we're trying to not grow much the # of UAST roles for the initial version and this would add two more.

proposals/bip-002.md
juanjux: I think the third one (not doing this) is implicit on any BIP :)

proposals/bip-002.md
```diff
@@ -0,0 +1,93 @@
...
+| Author | Juanjo Álvarez |
+| Status | Draft |
+| Created | 2017-06-01 |
+| Updated | 2017-06-01 |
+
+## Abstract
+
+This BIP proposed a new official (as in defined in the Node struct and
+documented) "AnnotationStatus" property that would mark if the role has been
+annotated in any way and the quality of the current annotation.
```
mcuadros: The same role in the same driver can be an incomplete annotation and a complete annotation?

juanjux: The complete / incomplete / un annotated status mark nodes, not roles (even tough the nodes could have roles). 

Example: a language has a normal function definition that could be mapped to FunctionDef without problems  but also some variant that is somewhat different (for example an async function def), since the async / await nodes are not in the UAST the same driver would annotate as "incomplete" only the second one.

proposals/bip-002.md
abeaumont: I don't understand, a node name should affect its annotations?
I was thinking about annotation related changes. I can see 3 cases:
- Annotations are removed. In this case until the driver is updated it will be reporting incorrect information.
- A better way to annotate is defined. In this case until the driver's annotation properties are updated, it will be reporting incorrect information (and it would require to update the driver just to make annotation status valid).
- New annotations are added that were previously unavailable. In this case, driver information will remain updated

proposals/bip-002.md
abeaumont: I was talking about removing *both* `Annotated` and `Unannotated` statuses, and just give the option to mark some nodes as `Incomplete`. If there's no `Incomplete` flag, then the node is either annotated or not depending on whether is actually annotated.

proposals/bip-002.md
abeaumont: If it's going to just be redundant information, I don't see the point of having it. See above.

proposals/bip-002.md
abeaumont: As I see it, the alternatives, with the proposal, should be the comprehensive set of all the possibilities. In many cases, not doing anything will not be an option since that would mean that Babelfish goals cannot be achieved. In that sense, I think this proposal is ortoghonal to Babefish goals, it may be a nice to have (or not), but not implementing it will not prevent us from reaching any specific goal.

proposals/bip-002.md
abeaumont: Hmmm, I fail to see how it's better to look to the AnnotationStatus property of a node instead of just the Roles property of the node, am I missing something?

juanjux: Easier to grep, easier to parse, but read below about the possible boolean "incomplete" property.

proposals/bip-002.md
juanjux: The annotations are going to be wrong either way if the UAST changes or remove something since the roles are an enum (that's why we need to add versions checking in the drivers). Being a property of the nodes will at least keep the "unannotated" and "incomplete" information unchanged.

proposals/bip-002.md
juanjux: You need to add a property to mark stuff as incomplete either way. It could be a tristate or it could be a boolean (I would prefer a tristate because I like explicit things) but if the other reviewers prefer a boolean that would serve equally. 

- States aren't going to be inconsistent since the Annotated or Unannotated values are added by the SDK automatically so, without bugs in the SDK , those states you mention should be impossible.

proposals/bip-002.md
juanjux: It's definitely a nice to have, not something critical for the future of the project.

proposals/bip-002.md
juanjux: PS: I'll add the option to only having an "Incomplete" mark in the BIP alternatives section.

proposals/bip-002.md
abeaumont: Yes, that's exactly my point, you may end up with a driver claiming nn% coverage when in fact their annotations are wrong. My original question asked about how is this expected to be handled. I infer you're suggesting some kind of version checking. How would that work?

juanjux: The % coverage would indicate how many nodes are annotated which is something. They could be wrong because of bugs in the annotation code, but checking that is outside the scope of this BIP and yes, we definitely need to add UAST version checking eventually (that has been already talked about previously). Anyway I'll add a comment about this in the BIP.

proposals/bip-002.md
abeaumont: I'm not discussing at this point which option is better, just enumerating valid options that I think should be considered, so it'll be great to see it included as an alternative or main proposal.
Having all the information explicit may be desirable, but that doesn't make having it explicitely _in the UAST spec_ mandatory. You can easily build a tool (a candidate for bblfsh/tools maybe?) or add a CLI parameter (a debug flag?) to generate a tree with all the explicit information with the debug information.

juanjux: Yeah that's the point of the BIPs, getting feedback, and I think the proposal of having only an incomplete tag/mark/role/whatever has merit so I'm adding it. 

I think that the tagging of incomplete roles should be mandatory to the driver programmers because it provides valuable information for the user. The Annotated/Unannotated information could be certainly generated by an SDK tool (also incorporating this into the BIP).

juanjux:
Feedback added into the BIP. Thanks!

proposals/bip-002.md
abeaumont: > The % coverage would indicate how many nodes are annotated which is something.

Not sure bad data is better than no data :)
Anyway, having proper versioning would really help, since it'd make easy to discard unmaintaned driver's coverage and it may also be an important metric by itself.

proposals/bip-002.md
```diff
@@ -63,6 +68,15 @@ roles. This would make this information available for potentials users that only
 check the UAST roles, but would increase the size of the UAST and open the door
 for more "meta" roles, so this author advises against it.
 
+Another proposed alternative would be to drop the "Unannotated" and "Annotated"
+statuses, since they can be inferred by the presence/ausence of roles in the Roles
```
abeaumont: *absence*

proposals/bip-002.md
```diff
@@ -63,6 +68,15 @@ roles. This would make this information available for potentials users that only
...
+field, and only use an "Incomplete" mark, either by using a boolean property in
+the struct or a new UAST meta-role. This would have the advantage of simplifying
+the implementation and reducing but would make the status less "greppable" and
```
abeaumont: something is missing after *reducing*

proposals/bip-002.md
```diff
@@ -63,6 +68,15 @@ roles. This would make this information available for potentials users that only
...
+less visually obvious when manually checking a generated UAST but that could also
+be done with an sdk tool that could then be called at the end of the integration
```
abeaumont: the tool may not be in the sdk, it could also be an external tool

proposals/bip-002.md
```diff
@@ -84,7 +98,9 @@ for more "meta" roles, so this author advises against it.
   `SetIncomplete()` on the appropriate nodes.
 
 - The integration-test implementation would have to be modified to provide the
-  summary of annotation completeness at the end of the run.
+  summary of annotation completeness at the end of the run. This basis for this
```
abeaumont: I don't understand what *This basis for this report...* means

proposals/bip-002.md
```diff
@@ -100,7 +100,9 @@ tests.
 - The integration-test implementation would have to be modified to provide the
```
abeaumont: This is just a use case, not sure that we need to specify how this annotation status will be used, how we implement this report of if we do it should be decided without having to alter this spec if we decide to do it differently. It may be interesting to add it as a use case in the rationale, though.

juanjux:
Closing this one, I'll open a new PR for discussing the second WIP version of the BIP including all the options given in the feedback of this PR.

PR #34, opened 2017-06-06
[getting started] Add a section about Babelfish Tools

abeaumont:


user/getting-started.md
```diff
@@ -85,6 +80,46 @@ $ echo "import foo" > sample.py
...
+- Get some data from the source code
+- See how a language analysis tool is implemented, as a basis for your own
+  tools.
+
+
+### Setup
+
+Running Babelfish Tools standalone requires getting the `bblfsh-tools`
+binary. Currently this requires a working setup of
+[Go](https://golang.org/). You can get it with the following command:
```
juanjux: Maybe this link is more direct for setting up Go: https://golang.org/doc/install

abeaumont: :+1: I took it from another paragraph in the same document, I'll change both

bzz:
Nice improvement!

@abeaumont while we are here, can we please clarify https://github.com/bblfsh/documentation/pull/34/files#diff-9b6a126e83a5470443b9f2399672749dR53

  > **[warning] Server only runs on Linux!**
  >
  > Babelfish Server relies on ..

server -> `single binary server` or `standalone server`. Otherwise it's a bit miss-leading for people who have skipped the upper part of docs on docker or might have landed here from search engine.

Also may be a nitpick, but `### Running with Docker (recommended)` and `#### Running standalone` may be better off having same number of `#`s.

From Topic of Content perspective, again may be not super important, but it's not immediately clear why `## Babelfish Tools` and `## Usage` have separate, level 2 headings. May be `### Usage` was meant here?

smola:
@bzz Technically Babelfish Server only runs on Linux, but Docker for Windows and Docker for Mac use a Linux VM to get it running. Do you think we really need to add this clarification?

bzz:
@smola up to you, it's just a suggestion! 

Babelfish is many things (SDK, server, tools, API, etc) and I see some value in making it clear for the first time user that Docker works anywhere, but single native binary for server works only for linux. 

Having few extra words in docs that might save some time for a user later on - sounds like a good investment. 

Second part about the ToC structure may be more relevant to take care of.

abeaumont:
Changes from received feedback applied

PR #36, opened 2017-06-06
Fixed links

juanjux:


bzz:
Looks good to me. 

Does anybody know if that something we have to do everywhere and doc generator is not going taking care of for us in foreseeable future? Asking as got used to other doc systems that do that i.e https://raw.githubusercontent.com/apache/spark/master/docs/cluster-overview.md

driver/annotations.md
```diff
@@ -2,11 +2,11 @@
 
 # Intro
 
-Once you have written the [AST parser](driver/sdk.md) the next step is to write
+Once you have written the [AST parser](https://doc.bblf.sh/driver/sdk.html) the next step is to write
```
smola: docbook does parse links such as `driver/sdk.md` and transforms them to the right links on building. The problem here is that links are always relative to the current file. So if you want to link `driver/sdk.md` from the `driver/annotations.md` file, you have to use `sdk.md` or `../driver/sdk.md`, but not `driver.sdk` (which produces a link to `driver/driver/sdk.md`).

smola:
@bzz Doc generator takes care (see my previous comment). Also, we now have verification in Travis that mark build as failed if there is any broken link (internal or external).

These broken links were introduced before we had link checking in Travis.

juanjux:
PR updated, links changed to relative paths.

PR #40, opened 2017-06-08
gRPC example page

juanjux:
This page explains a working example using the gRPC protocol to get the tokens of a Python helloworld.

### DIFF
--- a/user/server-grpc-example.md
+++ b/user/server-grpc-example.md
@@ -1,8 +1,9 @@
 # Example usage of the gRPC interface
 
-On the [previous section](server-protocol.md) we checked
-the format of the messages exchanged using the gRPC protocol. Here, we're
-going to put that knowledge into practice with a simple example usage of the Babelfish server using that protocol.
+On the [previous section](server-protocol.md) we checked the format of the
+messages exchanged using the gRPC protocol. Here, we're going to put that
+knowledge into practice with a simple example usage of the Babelfish server using
+that protocol.
 
 This example is done in Go but you could use the gRPC interface using [any
 programming language that supports this protocol](http://www.grpc.io/about#osp).
@@ -15,10 +16,10 @@ format](https://developers.google.com/protocol-buffers/docs/proto) defines the
 structures and methods used for protocol buffer data. In the case of the Babelfish
 server protocol, you need to get [this `.proto`
 file](https://github.com/bblfsh/sdk/blob/master/protocol/generated.proto) from the
-SDK. Then you have to compile it to a source file with the required structures
-and methods. How you generate this module is language-dependent but usually it
-involves installing gRPC using your language package management tool (pip, cargo,
-npm, dub, etc) and then use one of the provided tools.
+SDK. Then you have to compile it to a source file with the required structures and
+methods. How you generate this module is language-dependent but usually it
+involves installing gRPC using your language package management tool (if it has a
+package for it) and then use one of the provided tools.
 
 For this example, we'll use the [GogoProtobuf
 implementation](https://github.com/gogo/protobuf) for the Go language. This
@@ -88,7 +89,7 @@ The most important member of the `ParseUASTResponse` object is undoubtly
 detailed in the previous page](server-protocol.md#Nodes). This first node
 returned would be the root node of the UAST, and you typically would iterate over
 the node children (contained in the aptly named `children` field) typically using
-the [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) and 
+[a visitor](https://en.wikipedia.org/wiki/Visitor_pattern) and 
 reading the `token`s and `roles` in the tree to do your tool.
 
 For demonstration purposes, we'll just write a simple function that iterates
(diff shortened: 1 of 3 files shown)


user/server-grpc-example.md
```diff
@@ -0,0 +1,156 @@
...
+
+The first step involving gRPC communication is getting the `.proto` file that
+defines the types involved in the communication. The [`.proto`
+format](https://developers.google.com/protocol-buffers/docs/proto) defines the
+structures and methods used for protocol buffer data. In the case of the Babelfish
+server protocol, you need to get [this `.proto`
+file](https://github.com/bblfsh/sdk/blob/master/protocol/generated.proto) from the
+SDK. Then you have to compile it to a source file with the required structures
+and methods. How you generate this module is language-dependent but usually it
+involves installing gRPC using your language package management tool (pip, cargo,
```
abeaumont: minor: there's no gRPC for dub, and the cargo one is not recommended.

user/server-grpc-example.md
```diff
@@ -0,0 +1,156 @@
...
+as seen on the [server protocol](server-protocol.md) page. You should check
+the `status` (`Status` in the case of Go, since public members start with
+uppercase); only a value of `protocol.Status.OK` will indicate sucess.
+
+The most important member of the `ParseUASTResponse` object is undoubtly
+`uast` (`UAST` in Go). This will contain a `Node` object which the [structure
+detailed in the previous page](server-protocol.md#Nodes). This first node
+returned would be the root node of the UAST, and you typically would iterate over
+the node children (contained in the aptly named `children` field) typically using
+the [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) and
```
abeaumont: minor: design patterns jargon is not so used with higher level languages. Since this is language agnostic, I'd leave the visitor pattern reference out.

user/server-protocol.md
```diff
@@ -0,0 +1,66 @@
+# Babelfish Protocol
+
+The Babelfish client-server protocol uses [gRPC](http://www.grpc.io) for method
+selection (with the inherent Protobuf3 used for message exchanges). You can read
```
abeaumont: I find _the inherent Protobuf3 used_ confusing

user/server-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+files to see the format description of the messages and types involved, but we'll
+provide here a simple definition in JSON-like format. On the [next
+page](server-grpc-example.md) we'll see a demo of how this comes together in
+practice with some code.
+
+## ParseUASTRequest
+
+Issued by the client to request that a source code file must be parsed to 
+an UAST tree. The client must provide the code in the `content` field, the
+programming language in the `language` field (which obviously must be one
```
abeaumont: language may be empty for autodetection iirc

user/server-grpc-example.md
```diff
@@ -0,0 +1,156 @@
+# Example usage of the gRPC interface
+
+On the [previous section](server-protocol.md) we checked
```
juanjux: Works, because its in the same directory and they are relatives (to add the directory name it would have to be `../user/server-protocol.md`).

user/server-grpc-example.md
juanjux: True, I'll just remove the package managers and add "if it has a package for it".

user/server-grpc-example.md
juanjux: The visitor pattern is probably the most well known one after the singleton and factory even on languages like Python so I don't think many programmers will miss the reference.

user/server-protocol.md
juanjux: It is, I'll reword it, thanks.

user/server-protocol.md
juanjux: Didn't knew that, adding info, thanks.

user/server-grpc-example.md
abeaumont: Heh, I should problably have defined what I meant by _higher level_ better, certainly Python is not a language I'd include in. Anyway, the point is that _design patterns_ are highly opinionated and that's why it may make less sense to many people.

If you want to keep it, there's now  a _the_ that should be removed

juanjux: Higher level = functional or what? Ooook...haha.

I don't see how the concept of pattern can be opinionated; they're generic recipes for some  generic problems, you can argue that there are programmers that are talibanish about them (aka MacroProgrammers) to the point of not thinking solutions by themselves, which I think is true but the concept of pattern is very clearly understood among most programmers (maybe not higher level ones tough). 

Fixed the spurious "the", thanks.

abeaumont: Hmmm, yes, functional language would probably have been a better name. One could argue that most design patterns (as we know them) are simplified or just simply disappear with first class functions, and that design patterns were just build for languages lacking that features. That's why I consider less universal than usually considered.

juanjux: Actually, all functions (and even types and modules) are first class objects in Python :)

Certainly some patterns can be better implemented when you have features in your arsenal like first class functions (or code generation), but that doesn't mean they aren't useful as nomenclature for referring to certain types of implementations (like a visitor in this case).

abeaumont: I know, yet functional programming is not considered _pythonic_ and scarcely used. And sure, they can be useful as nomenclature, I just find it more universal to talk about tree traversal than about visitor pattern

PR #53, opened 2017-06-28
fix error handling example

smola:


bzz:
Looks good to me.

user/server-grpc-example.md
```diff
@@ -144,6 +145,11 @@ func main() {
 	if err != nil {
 		os.Exit(1)
 	}
+
+	if resp.Status != protocol.Ok {
```
abeaumont: Since this example is for illustrative purposes, it may be interesting to add an error message per error to better explain what the problem is. In this case something like
_Source code has not been parsed correctly_, or similar.

smola: Updated it to include an error message, also making it explicit that `resp.Errors` might contain multiple errors.

PR #56, opened 2017-06-30
bump driver versions

smola:


smola:
Merge only once images are in Docker Hub.

PR #64, opened 2017-07-14
documented --max-message-size option

mcarmonaa:


mcarmonaa:
Since I can't assign reviewers yet, I mention to mantainer: @juanjux

user/getting-started.md
```diff
@@ -120,6 +120,20 @@ DEBU[0000] registering gRPC service
 INFO[0000] starting gRPC server
 ```
 
+
+### Setting maximum message size
+
+If a customized [gRPC](https://grpc.io) message size is needed, you can use the command flag option `--max-message-size`. By default  [gRPC](https://grpc.io) uses 4MB as the upper limit if any other value is set, but you can override it:
```
juanjux: I think the `if any other value is set` should be `if any other value is not set` but it would be better removed since that's implicit in the concept of default.

bzz:
@mcarmonaa does the documentation for Docker works on your machine?

On my env it results in 

```
docker run --privileged -p 9432:9432 --name bblfsh bblfsh/server:  --max-message-size=100

docker: Error parsing reference: "bblfsh/server:" is not a valid repository/tag: invalid reference format.
See 'docker run --help'.
```

PR #71, opened 2017-08-01
New rewritten BIP

juanjux:
This is not a BIP but something like a "WIP" BIP. It provides options for the alternatives proposed in the previous PR and the new ones given from the feedback, and weights advantages and disadvantages to each. The purpose of this PR is to discuss the WIPBIP in it's current form and for the Benevolent Dictator @mcuadros to choose it's own adventure based on the alternatives provided (or the new ones from this PR discussion).

Once that's done I would do a new PR with the BIP in it's final form ready to be included if approved.

### DIFF
--- a/proposals/bip-002.md
+++ b/proposals/bip-002.md
@@ -5,15 +5,15 @@
 | BIP | 2 |
 | Title | Explicit Annotated status and completeness information |
 | Author | Juanjo Álvarez |
-| Status | Draft |
+| Status | Accepted |
 | Created | 2017-06-01 |
-| Updated | 2017-08-01 |
+| Updated | 2017-08-02 |
 
 ## Abstract
 
-This BIP proposed a new official (as in defined in the Node struct and
-documented) "AnnotationStatus" property that would mark if the role has been
-annotated in any way and the quality of the current annotation.
+This BIP proposed two new official `Incomplete` and `Unannotated` roles would mark
+if the node has been annotated in any way and the quality of the current
+annotation.
 
 ## Rationale
 
@@ -26,11 +26,8 @@ This is a worthy objective but has the problem of leaving the nodes related with
 these features unannotated or with annotations that don't fully express all
 the semantic information that the language source code or the native AST have.
 
-This WIP-BIP proposes a mechanism to mark nodes with completion status (which
-could use properties or roles and could vary on the mark details, see below in
-*Proposed implementations* section).
-
-This will make the UAST user have a wider view of the completeness of the
+This BIP proposes a mechanism to mark nodes with two possible completion status
+roles. This will make the UAST user have a wider view of the completeness of the
 information provided by it on specific nodes and also could enable new uses cases
 (whose specific definition is outside of the scope of this BIP) like helping debug
 drivers and provide a completeness summary, with the added benefit of providing a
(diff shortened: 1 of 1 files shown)


abeaumont:
As talked in person, we've decided to:

* There will be an `Incomplete` role that the driver developer can use to mark roles.
* There will be an `Unannotated` role that will be set by default.
* The sdk will provide a tool to say which roles are currenly unannotated, but there may be unannotated roles that the driver doesn't know about (for example, if the driver doesn't support new language features).

proposals/bip-002.md
```diff
@@ -0,0 +1,99 @@
...
+| BIP | 2 |
+| Title | Explicit Annotated status and completeness information |
+| Author | Juanjo Álvarez |
+| Status | Accepted |
+| Created | 2017-06-01 |
+| Updated | 2017-08-02 |
+
+## Abstract
+
+This BIP proposed two new official `Incomplete` and `Unannotated` roles would mark
```
abeaumont: not sure `official` is necessary here, there're no `unofficial` roles :)

juanjux:
BIP updated, options replaced by the choosen ones, made it a little shorter while I was at it.

proposals/bip-002.md
```diff
@@ -0,0 +1,99 @@
...
+drivers and provide a completeness summary, with the added benefit of providing a
+numerically observable measure of the completeness of every developed driver.
+
+## Example
+
+If a version of the UAST specification didn't have a role that could be matched
+for a "coroutine" type function supported by some language, the UAST could still
+mark those nodes as "FunctionDef", but without additional information, this would
+be providing incomplete and even misleading information to UAST users that could
+treat coroutines differently (for example when making stack call deep analysis).
```
abeaumont: * s/deep/depth/?
* The example is not very clear... a simpler one may be better, idk

proposals/bip-002.md
```diff
@@ -0,0 +1,99 @@
...
+to these coroutine nodes and the UAST user for that language could then look at
+the native AST properties to complete their semantic understanding of the node.
+
+## Specification
+
+### UAST Roles
+
+This change will add two new roles the UAST definition:
+
+- `Incomplete`: For nodes that have been annotated annotations doesn't express the
```
abeaumont: * s/doesn't/don't/
* is some word missing between _annotated_ and _annotations_?

proposals/bip-002.md
```diff
@@ -0,0 +1,99 @@
...
+- We'll add a new process to the build system that will check both the
+  `driver/normalizer/??ast/??ast.go` file with the static list of nodes for
+  existing but unannotated nodes and the integration test results and output a
+  summary of incomplete an unannotated nodes.
+
+- The documentation for the driver developer would have to be updated to include 
+  the purpose and mechanics of these new property/s or role/s. This is in the file 
+  `driver/annotations.md`.
+
+- This BIP will be added to the official documentation in the `proposals`
```
abeaumont: This is independent of whether it's approved or not, so I don't think it's an impact of approving it.

juanjux: Maybe not added by *published* (like in showing in http://doc.bblf.sh), will change it.

proposals/bip-002.md
juanjux: True.

proposals/bip-002.md
juanjux: I will reword it in a shorter form using list comprehensions.

proposals/bip-002.md
juanjux: - Fixed
- Yes. Added *but the*

juanjux:
@abeaumont discarded alternatives section added.

abeaumont:
@juanjux I think the alternatives section lack more information. The selection was done for a reason and alternatives were also discarded for a reason. A reason that we have clear now but that we won't have when we (or any other newcomer) look back at this BIP in the future and ask why this selection was done instead of any of the alternatives, unless we make it explicit now in the document.

PR #74, opened 2017-08-04
Added clarification about Incomplete roles alone

juanjux:
- Added an extended description of the `Incomplete` role.
- Formatting fix.

### DIFF
--- a/proposals/bip-002.md
+++ b/proposals/bip-002.md
@@ -46,12 +46,11 @@ marking the incompleteness of this information to the UAST user.
 
 This change will add two new roles the UAST definition:
 
-- `Incomplete`: For nodes that have been annotated but the annotations don't
-  express the full semantic meaning or there aren't any other annotations because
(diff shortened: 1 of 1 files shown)


proposals/bip-002.md
```diff
@@ -47,7 +47,9 @@ marking the incompleteness of this information to the UAST user.
 This change will add two new roles the UAST definition:
 
 - `Incomplete`: For nodes that have been annotated but the annotations don't
-  express the full semantic meaning.
+  express the full semantic meaning or there aren't any other annotations because
```
abeaumont: A bit shorter and more clear maybe:
_For nodes that don't have their full semanting meaning annotated, including both partially annotated nodes and nodes without any other annotation_

juanjux: I like it. I tend to be too verbose, yes. Also reworded the `Unannotated` description a little:

For nodes that have received no other annotations. Automatically
 added by the SDK. Driver's developers should aim to have none of these.*

Issue #76, opened 2017-08-09
Document that it's neccesary to share `docker.sock` for the docker-transport option to work


juanjux:

juanjux:
Fixed by #77.

PR #77, opened 2017-08-09
Added reference to share the docker.sock when using the local docker daemon for images

juanjux:


juanjux:
Fixes #76.

user/getting-started.md
```diff
@@ -107,6 +110,11 @@ time="2017-07-12T14:11:13Z" level=debug msg="starting server"
 time="2017-07-12T14:11:13Z" level=debug msg="registering gRPC service"
 time="2017-07-12T14:11:13Z" level=info msg="starting gRPC server"
 ```
+
+Notice how in this case we need to share the host Docker server's Unix socket with
```
abeaumont: I think there's a misconception here, or at least it's not clear enough to me with current explanation. With this example, it seems that you always need to share your docker daemon socket, and that you need to use `docker-daemon` transport to get your images.

This is neither necessary not the best option. The simplest way to use your custom images is to just let the server fetch the custom images from dockerhub instead:
```
$ BBLFSH_DRIVER_IMAGES="python=docker://bblfsh/python-driver:dev-136e550" docker run -e BBLFSH_DRIVER_IMAGES --privileged -p 9432:9432 --name bblfsh bblfsh/server
```

Using docker-daemon is just an additional, more complex way to get the same result that only makes sense if you already have the image handled by your docker (you have built it or similar). But this is a more advanced approach.

juanjux: Well, the objective of this PR wasn't an overwrite of the "Overriding driver images section" but since I actually agree that the easiest option is not shown I've changed it.

user/getting-started.md
```diff
@@ -95,9 +95,19 @@ In case you need that the Babelfish server to run different driver images than t
 
     BBLFSH_DRIVER_IMAGES="language=docker-registry:namespace/repository:tag;language2=docker-registry:namespace/repository:tag"
 
-So if you are running the server with a Docker container you could do something similar to:
+So to get a specific version of an image from
+[Dockerhub](https://hub.docker.com/u/bblfsh/) the line would be:
 
+```bash
+$ BBLFSH_DRIVER_IMAGES="python=docker://bblfsh/python:v0.4.2" docker run \
```
abeaumont: It's not `python:v0.4.2` but `python-driver:v0.4.2`.

user/getting-started.md
```diff
@@ -95,9 +95,19 @@ In case you need that the Babelfish server to run different driver images than t
...
+  -e BBLFSH_DRIVER_IMAGES -v /var/run/docker.sock:/var/run/docker.sock \
```
abeaumont: Sharing `docker.sock` is not needed in this case.

user/getting-started.md
```diff
@@ -95,9 +95,19 @@ In case you need that the Babelfish server to run different driver images than t
...
+  --privileged -p 9432:9432 --name bblfsh bblfsh/server
 ```
+
+Instead, if you want the server to retrieve specific driver images from a local Docker
+daemon (e.g. when testing a driver that you're developing) you could do something similar to:
+
+```bash
 $
 BBLFSH_DRIVER_IMAGES="python=docker-daemon:bblfsh/python:dev-96b24d3;java=docker-daemon:bblfsh/java-driver:latest"
```
abeaumont: there's an inconsistency here between `python` and `java-driver`

juanjux:
Fixes applied, thanks.

PR #82, opened 2017-09-04
proposals: Add new BIP-003 proposal, Agglutinative Roles language

abeaumont:
Part of https://github.com/bblfsh/sdk/issues/167

### DIFF
--- a/proposals/bip-003.md
+++ b/proposals/bip-003.md
@@ -1,0 +1,460 @@
+# Agglutinative Roles language
+
+| Field | Value |
+| --- | --- |
+| BIP | 3 |
+| Title | Agglutinative Roles language |
+| Author | Alfredo Beaumont |
+| Status | Draft |
+| Created | 2017-09-04 |
+| Updated | 2017-09-04 |
+
+## Abstract
+
+Current Role set includes roles that contain multiple properties,
+instead of having a single Role per property.
+
+We propose modifying current role set to include only roles that represent a single property
+and to consider the Role set an agglutinative language,
+annotating each node with a set of Roles to fully express node properties.
+
+We show how this can make the Role vocabulary more scalable, flexible and expressive,
+and how it can make the work of code analysts easier.
+
+## Rationale
+
+In current Role set many roles contain multiple properties (i. e.: `OpPreIncrement` for `++i`).
+
+This presents some issues:
+* It doesn't scale well.
+  For a set of N properties, in the worst case 2^N roles would be needed.
+* It makes code analysis hard.
+  For example, to search for all the operators you need to search for the union of many roles.
+* There's hard to provide partial information for rare properties,
+  if some property is not common enough as to deserve its own role.
+
+A more flexible and scalable way to define the role set would be to have one role per property
+and to allow to combine these roles/properties as needed.
+In the example above `OpPreIncrement` would become the union of 3 roles (`Operator`, `Increment`, `Prefix`).
+
+This combination of roles is already done to some extent,
+since a preincrement operator would actually be annotated with 2 roles: `Expression`, `OpPreIncrement`.
+Current proposal just deepens this property separation into roles.
+
+As a derivative advantage of this new role language and its scalability,
+new roles could be defined that better capture some additional property shared by different tokens
+that will make code analysis even easier.
+
+Following the previous example, we could use some additional roles, like `Arithmetic`, `Unary`, etc.
+to express additional properties of the role.
+So the final annotation could be something like: `Operator`, `Arithmetic`, `Unary`
+
+* `Expression` (or `Statement`, depending on the language)
+* `Operator`
+* `Arithmetic`
+* `Increment`
+* `Unary`
+* `Prefix`
+
+With this kind of classification a code analyst that may be interested in analysing the usage of, say,
+arithmetic operators, would have an easier way to filter the `UAST` to find the interesting nodes.
+
+Additionally, this agglutination of roles makes unsupported node types degrade more gracefully.
+For example, a `log` operator, which currently lacks an specific role,
+would be just left as: `Expression`, `Incomplete`.
+With the new language, the node could still retain most of the information:
+
+* `Expression`
+* `Operator`
+* `Arithmetic`
+* `Binary`
+
+So can extract that it's a binary arithmetic operator generically,
+restricting the need to directly access the token to only those analysis
+interested in this kind of tokens specifically.
+
+## Specification
+
+The set of Roles are changed by this proposal.
+It's limited to partition the multiple property roles currently defined,
+leaving the potential addition of new property roles
+(`Arithmetic`, `Comparsion`, `Loop`, ...) to a future BIP.
+
+The following roles are proposed for removal:
+
+* `SimpleIdentifier`
+* `QualifiedIdentifier`
+* `BinaryExpression`
+* `BinaryExpressionLeft`
+* `BinaryExpressionRight`
+* `BinaryExpressionOp`
+* `OpBitwiseLeftShift`
+* `OpBitwiseRightShift`
+* `OpBitwiseUnsignedRightShift`
+* `OpBitwiseOr`
+* `OpBitwiseXor`
+* `OpBitwiseAnd`
+* `OpEqual`
+* `OpNotEqual`
+* `OpLessThan`
+* `OpLessThanEqual`
+* `OpGreaterThan`
+* `OpGreaterThanEqual`
+* `OpSame`
+* `OpNotSame`
+* `OpContains`
+* `OpNotContains`
+* `OpPreIncrement`
+* `OpPostIncrement`
+* `OpPreDecrement`
+* `OpPostDecrement`
+* `OpNegative`
+* `OpPositive`
+* `OpBitwiseComplement`
+* `OpDereference`
+* `OpTakeAddress`
+* `OpBooleanAnd`
+* `OpBooleanOr`
+* `OpBooleanNot`
+* `OpBooleanXor`
+* `OpAdd`
+* `OpSubstract`
+* `OpMultiply`
+* `OpDivide`
+* `OpMod`
+* `PackageDeclaration`
+* `ImportDeclaration`
+* `ImportPath`
+* `ImportAlias`
+* `FunctionDeclaration`
+* `FunctionDeclarationBody`
+* `FunctionDeclarationName`
+* `FunctionDeclarationReceiver`
+* `FunctionDeclarationArgument`
+* `FunctionDeclarationArgumentName`
+* `FunctionDeclarationArgumentDefaultValue`
+* `FunctionDeclarationVarArgsList`
+* `TypeDeclaration`
+* `TypeDeclarationBody`
+* `TypeDeclarationBases`
+* `TypeDeclarationImplements`
+* `VisibleFromInstance`
+* `VisibleFromType`
+* `VisibleFromSubtype`
+* `VisibleFromPackage`
+* `VisibleFromSubpackage`
+* `VisibleFromModule`
+* `VisibleFromFriend`
+* `VisibleFromWorld`
+* `IfCondition`
+* `IfBody`
+* `IfElse`
+* `SwitchCase`
+* `SwitchCaseCondition`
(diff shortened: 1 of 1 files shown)


proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+We show how this can make the Role vocabulary more scalable, flexible and expressive,
+and how it can make the work of code analysts easier.
+
+## Rationale
+
+In current Role set many roles contain multiple properties (i. e.: `OpPreIncrement` for `++i`).
+
+This presents some issues:
+* It doesn't scale well.
+  For a set of N properties, in the worst case 2^N roles would be needed.
```
juanjux: Missing comma after case.

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+* There's hard to provide partial information for rare properties,
+  if some property is not common enough as to deserve its own role.
+
+A more flexible and scalable way to define the role set would be to have one role per property
+and to allow to combine these roles/properties as needed.
+In the example above `OpPreIncrement` would become the union of 3 roles (`Operator`, `Increment`, `Prefix`).
+
+This combination of roles is already done to some extent,
+since a preincrement operator would actually be annotated with 2 roles: `Expression`, `OpPreIncrement`.
+Current proposal just deepens this property separation into roles.
```
juanjux: The current proposal.

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+* `Arithmetic`
+* `Increment`
+* `Unary`
+* `Prefix`
+
+With this kind of classification a code analyst that may be interested in analysing the usage of, say,
+arithmetic operators, would have an easier way to filter the `UAST` to find the interesting nodes.
+
+Additionally, this agglutination of roles makes unsupported node types degrade more gracefully.
+For example, a `log` operator, which currently lacks an specific role,
```
juanjux: Nitpicking: at first I tough about logging, maybe change for `natural logarithm` operator?

juanjux: lacks *a* specific

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+So can extract that it's a binary arithmetic operator generically,
+restricting the need to directly access the token to only those analysis
+interested in this kind of tokens specifically.
+
+## Specification
+
+The set of Roles are changed by this proposal.
+It's limited to partition the multiple property roles currently defined,
+leaving the potential addition of new property roles
+(`Arithmetic`, `Comparsion`, `Loop`, ...) to a future BIP.
```
juanjux: `Comparison`

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+  but can only go with a `noun`.
+  We consider that this may be a valid approach,
+  but it can be left for consideration in the future,
+  after more experimentation,
+  since it'd make the proposal more complex while not having a clear gain
+  and it's easy to extend in that direction in the future.
+
+## Impact
+
+Imcompatible changes to the Role set are proposed, in order to do that,
```
juanjux: Incompatible

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+the following changes would be needed:
+
+* Versioning should be added to the [SDK](https://github.com/bblfsh/sdk/),
+  to allow existing server and drivers work with a previous version of the SDK.
+* Roles should be updated in the SDK.
+* Protobuf generated code should be updated for [server](https://github.com/bblfsh/server) and [python](https://github.com/bblfsh/client-python) and [go](https://github.com/bblfsh/client-go) clients
```
juanjux: Python and Go (uppercase).

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+* `FunctionDeclarationArgumentDefaultValue`: `Function`, `Declaration`, `Argument`, `Value`
+* `FunctionDeclarationVarArgsList`: `Function`, `Declaration`, `Argument`, `ArgsList`
+* `TypeDeclaration`: `Type`, `Declaration`
+* `TypeDeclarationBody`: `Type`, `Declaration`, `Body`
+* `TypeDeclarationBases`: `Type`, `Declaration`, `Base`
+* `TypeDeclarationImplements`: `Type`, `Declaration`, `Implements`
+* `VisibleFromInstance`: `Visibility`, `Instance`
+* `VisibleFromType`: `Visibility`, `Type`
+* `VisibleFromSubtype`: `Visibility`, `Subtype`
+* `VisibleFromPackage`: `Visiblity`, `Package`
```
juanjux: Visibility

juanjux:
Good job!

About possible improvements, I miss a note in the `Impact` section about the current users of the UAST having to update their code and the way they interpret the roles.

Also, a third alternative could be the "two-level UAST" I proposed some time ago, where the first level roles are the more generic (`loop`, `branch`, `procedure`) and a second level would have a more concrete semantic meaning (`foreach`, `if`, `coroutine`). But it is probably tangential for this proposal (both levels could use agglutinative roles) so it doesn't need to be added.

Other than that, this BIP has my ACK for the main proposal.

proposals/bip-003.md
abeaumont: Changed to `**` (pow), to avoid potential confusion.

abeaumont:
Updated with typo and grammar fixes from the review.

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+* `Substract`
+* `Multiply`
+* `Divide`
+* `Modulo`
+* `Package`
+* `Declaration`
+* `Import`
+* `Path`
+* `Alias`
+* `Function`
```
mcuadros: we don't have a Class role?

proposals/bip-003.md
```diff
@@ -0,0 +1,460 @@
...
+
+With this kind of classification a code analyst that may be interested in analysing the usage of, say,
+arithmetic operators, would have an easier way to filter the `UAST` to find the interesting nodes.
+
+Additionally, this agglutination of roles makes unsupported node types degrade more gracefully.
+For example, a `**` (pow) operator, which currently lacks a specific role,
+would be just left as: `Expression`, `Incomplete`.
+With the new language, the node could still retain most of the information:
+
+* `Expression`
```
ajnavarro: In this or other cases, can be a possibility that a role with specific properties can match one by one arguments from another role and be nodes slightly different?

proposals/bip-003.md
juanjux: That would be `TypeDeclaration` or `Type+Declaration` with the new roles.

proposals/bip-003.md
abeaumont: For nodes with incomplete roles, that can certainly be the case. You could have unsupported operators (let's say `pow` and `log`), that you know they're arithmetic operators, but cannot distinguish between them. You'd have to go and check the token, if available, in that case.

eiso:
First of all, thank you for the work on bip-3. It's a really interesting and very well written proposal. Having been giving it some thought today it makes a lot of sense to me (de)composing the roles, for the reasons that you state. 

The one concern that I have, is around higher level abstractions on the roles for easy usability (what @juanjux I believe calls 2 level roles), in particular in the [spark-api](https://github.com/src-d/spark-api) project. Since that project will be the way a large # of our intended audience will use babelfish, with the composed roles I can imagine some difficulties. 

Here is some pseudo code from the Spark API design document.

```Python
//DS - PySpark
//for 1000 Python repositories (repos.txt or LanguageDataset demo)
src-d.select()

//clone to local FS (or use .siva files files in hdfs://)
src-d.clone(repos.txt, “/path/to/cloned/repos”)

//get UASTs for HEAD
uasts = src-d.read.gitLocal(“/path/to/cloned/repos”)
  .getReferences().filter($“reference_name” === “HEAD”)
  .getFiles()
  .applyEnry()
  .applyBblfsh()

//extract SimpleIdentifier roles (ids)
uasts.filter("uast uast_lib('//*[@class=SimpleIdentifier]')")
```

`uast_lib` is essentially libuast as you can see here. If instead of being able to extract all simple identifiers, I know need a rule based system for roles (e.g. include x, but excludes y, z). I can imagine the syntax becoming difficult, less easy to get started with and also more intense to process on the Spark side. That's why I am tagging @ajnavarro @bzz and @erizocosmico here as well.

The above could be partially solved with the two-level UAST suggestion of @juanjux .

abeaumont:
@eiso that is a very valid concern, and I see now that I forgot to give a proper answer to @juanjux suggestion, sorry.
I've two comments on this topic:
1) I think what both @juanjux and @eiso are suggesting is just a particular case of the alternative approach I included at https://github.com/bblfsh/documentation/pull/82/files#diff-b3182c12132bb09dddfe38e712ae248bR289, which just talks about a categorization in general. The current proposal doesn't deepen in that approach for the reasons commented there, but if you consider otherwise, we can go deeper into that.
2) It's true that with this proposal search may become more complex but apart from categorization there are two additional tools we can use to solve them:
- Use appropiate roles to facilitate code analysis. It may be true that a code analyst may want to look for simple identifiers and doesn't want to get the qualified identifiers. So let's say that's the case for discussion's sake. Then we could add a `Simple` role to handle this properly and make analyst's life easier. I think the new approach makes this easier. Look for example at the arithmetic operator example I used, it would make it easier to look for arithmetic operators instead of all of them. The same way we can add roles for say, relational operators, control flow constructs, looping constructs, etc.
- Use the power of `xpath` syntax to do the search. Even if we missed some valid searches needed by code analysts with our roleset, I think it would still be a reasonable approach (at the usability level) to use something like: `//*[@roleIdentifier and not(@roleQualified)]`. Not sure about performance, we may need to check that.

Note that these points are presented to have a wider view of the possibilities, not to discourage a categorization (of two levels or otherwise), which I consider a valid approach.

eiso:
@abeaumont regarding topic 1, could you take your suggested approach and have a 'category type' role that gets added in the same manner. Or would this be mixing concepts?

abeaumont:
@eiso I'm not sure I understand what you mean, do you mean having a special field/attribute in a node named `type` which would contain the main category of a node? Something like:

```yaml
internalType: SimpleIdentifier
type: Identifier
roles: Simple, OtherRole, ...
```

If that's what you mean, yes, that could be a way to do it. If not, please elaborate a bit on your appoach.

eiso:
@abeaumont the options I meant were:

Option 1. When `ForEach` becomes `For`, `Iterator`, adding `Loop` as a sub-role. So having (`For`, `Iterator`, `Loop`). Where `Loop` is defined in the spec as a higher level role.

Option 2. ...or having (`For`, `Iterator`, `Loop`) but typed as:

```
role: For
type: level2

role: Iterator
type: level2

role: Loop
type: level`
`

Please ignore the terrible naming.

**Comment by abeaumont** [comment]
@eiso ok, I understand now. I think both options would be similar from a Babelfish point of view. I think we could add automatic support for option 2, to make role 'level' explicit, without much work, it would just be more verbose.

So the main question would be if this approach would be of any use for code analysis. I think an analyst would need to know the roles and their categories beforehand anyway, but you surely have a better code analysis perspective and I guess you may have some use case in mind where this kind of annotation would be of help?

File: proposals/bip-003.md [hunk 10]
```diff
@@ -0,0 +1,460 @@
...
+* `PackageDeclaration`: `Package`, `Declaration`
+* `ImportDeclaration`: `Import`, `Declaration`
+* `ImportPath`: `Import`, `Path`
+* `ImportAlias`: `Import`, `Alias`
+* `FunctionDeclaration`: `Function`, `Declaration`
+* `FunctionDeclarationBody`: `Function`, `Declaration`, `Body`
+* `FunctionDeclarationName`: `Function`, `Declaration`, `Identifier`
+* `FunctionDeclarationReceiver`: `Function`, `Declaration`, `Receiver`
+* `FunctionDeclarationArgument`: `Function`, `Declaration`, `Argument`
+* `FunctionDeclarationArgumentName`: `Function`, `Declaration`, `Argument`, `Name, `
```
juanjux: `,` at the end.

PR #91, opened 2017-09-20
Upgrade documentation

abeaumont:
Changes to reflect usage of SDKv1, updated roles, updated status of projects, etc.

### DIFF
--- a/user/language-clients.md
+++ b/user/language-clients.md
@@ -22,25 +22,30 @@ for these languages:
 The client API's differ to adapt to their language specific idioms, the following
 code shows a simple example with the Python client:
 
-```python
-from bblfsh import BblfshClient
-# Protobuf definition file:
-from github.com.bblfsh.sdk.uast.generated_pb2 import Node
-from bblfsh.launcher import ensure_bblfsh_is_running
+```go
+package main
 
-if __name__ == "__main__":
-    # Create a Babelfish server instance if it's not already running:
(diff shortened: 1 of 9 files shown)


user/language-clients.md
```diff
@@ -22,25 +22,30 @@ for these languages:
 The client API's differ to adapt to their language specific idioms, the following
 code shows a simple example with the Python client:
```
juanjux: Changed to code from Python to Go (why?) but not this sentence ("with the Python client"). Would be nice to add a `filter(query)` call too.

abeaumont: Oh, I should had elaborated on this change. I changed it mostly because upgrading python code meant that it got more complicated imports:

```python
ProtocolServiceStub = importlib.import_module(
    "bblfsh.gopkg.in.bblfsh.sdk.%s.protocol.generated_pb2_grpc" % VERSION
).ProtocolServiceStub
```

Which looks rather complex for an example. It also depends on the client's grpc files.
So I rewrote it in Go to keep it simpler.

The reason for not adding a `filter` call is mainly that I think we need a whole new section related to `xpath` and queries, so I didn't want to add a single sample line without further explanation and I didn't want to include all the `xpath` related changes in this update.

juanjux:
Travis is failing, seems to be related to some external connection problem, I'll retry a couple times more and merge anyway if it continues failing.

PR #104, opened 2017-10-26
Update the integration tests' information which the latest changes.

juanjux:

### DIFF
--- a/driver/sdk.md
+++ b/driver/sdk.md
@@ -316,7 +316,7 @@ difference is found.
 
 For this to work, the driver developer will need to provide files with source
 code examples in the `tests/` directory. with the `.source` file extension. It's
-recomended that you keep the original extension before the `.source` so the
+recommended that you keep the original extension before the `.source` so the
 command below can autodetect the language.
 
 Once this is done, you need to generate the `.native` and `.uast` fixtures with 
@@ -346,11 +346,11 @@ make integration-test
 
 This will generate the `.native` and `.uast` files from the driver in the 
 current directory and compare them with the previously generated files, failing 
-and printing a diff if it detect any changes. 
+and printing a diff if it detects any changes. 
 
 If you need to regenerate some of the files for a new version that changes the 
 output of the native AST or the UAST, you'll need to regenerate the affected 
-files **with the current version installed in the server**. This can be done 
+files **with the current version installed on the server**. This can be done 
 with the commands:
 
 ```
--- a/driver/sdk.md
+++ b/driver/sdk.md
@@ -311,24 +311,59 @@ change it since it'll execute the driver.
 
 The integration tests will test all the process of the driver from the request
 to the parser to the annotated UAST generation. They work by comparing the
-UAST output with the previously existing one, and will fail if any difference
-is found.
+UAST and native output with the previously existing one, and will fail if any 
+difference is found.
 
 For this to work, the driver developer will need to provide files with source
-code examples in the `tests/` directory. with the `.source` file extension.
-Once this is done, you need to run the integration-test Makefile's target for
-the first time to automatically generate  the `.native` and `.uast` files:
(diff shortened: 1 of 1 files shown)


driver/sdk.md
```diff
@@ -311,24 +311,59 @@ change it since it'll execute the driver.
...
-UAST output with the previously existing one, and will fail if any difference
-is found.
+UAST and native output with the previously existing one, and will fail if any 
+difference is found.
 
 For this to work, the driver developer will need to provide files with source
-code examples in the `tests/` directory. with the `.source` file extension.
-Once this is done, you need to run the integration-test Makefile's target for
-the first time to automatically generate  the `.native` and `.uast` files:
+code examples in the `tests/` directory. with the `.source` file extension. It's
```
abeaumont: the dot after `directory` is not needed.

driver/sdk.md
```diff
@@ -311,24 +311,59 @@ change it since it'll execute the driver.
...
+```
+# Only if you didn't already had a bblfshd server running:
+docker run -d --name bblfshd -p 9432:9432 bblfsh/bblfshd
+# Only if you didn't already had the drivers installed:
+docker exec -it bblfsh bblfshctl driver install --all
+
+bblfsh-sdk-tool fixtures fixtures/*.source
+```
+
+If the command fails with the error `unexpected error: runtime failure: missing
```
abeaumont: I don't think this makes sense. If you already have installed all the drivers you shouldn't see this problem, but if you see this error how can passing the language explicitely help with that?

driver/sdk.md
```diff
@@ -311,24 +311,59 @@ change it since it'll execute the driver.
...
-manually check carefully the `.uast` files because they'll be used as reference on
-next runs. Once you're happy with them, any subsequent executions of the
-`make integration-test` will actually compare them.
+This will generate the `.native` and `.uast` files from the driver in the 
+current directory and compare them with the previously generated files, failing 
+and printing a diff if it detects any changes. 
+
+If you need to regenerate some of the files for a new version that changes the 
+output of the native AST or the UAST, you'll need to regenerate the affected 
+files **with the current version installed on the server**. This can be done
```
abeaumont: This should work with any version of the server unless the `SDK` changes, shouldn't it?

driver/sdk.md
```diff
@@ -311,24 +311,59 @@ change it since it'll execute the driver.
...
+files **with the current version installed on the server**. This can be done 
+with the commands:
+
+```
+make build # build the driver docker image
+bblfsh driver remove mylang
+docker images
+# (…check the output to get the tag of the just build driver)
+bblfsh driver install mylang 
+docker-daemon=bblfsh/mylang-driver:dev-12345abc-dirty
```
abeaumont: Shouldn't it be `:` instead of `=`?

driver/sdk.md
juanjux: This error happens not only when there is not a driver but also when enry fails to detect the language, usually because it doesn't have the extension (like in the case of the Java fixtures, trough I'll make a PR about them). I'll be more explicit about this.

abeaumont: I'd say that this is then a bug in the server. The server shouldn't return this error when `enry` fails to detect the language.

driver/sdk.md
juanjux: It referred to the driver version (the one in development, since it's in the "developing drivers" page), but re-reading it I understand it could be confusing, I'll be more explicit here, thanks.

driver/sdk.md
juanjux: Yes. The problem is that enry.GetLanguage() only returns an empty string if it can't detect anything and then the server searchs for a driver called `""` which obviously doesn't work. It should detect the empty string and produce an specific error (I'll make a server PR update the doc).

PR #110, opened 2017-11-07
Added information about running a single driver container.

juanjux:


user/getting-started.md
```diff
@@ -180,3 +180,25 @@ Now you can add drivers from the local Docker using `bblfshctl` like this:
...
+To run a driver independently of the server run:
+
+```bash
+docker run --rm -p 9432:9432 bblfsh/python-driver
+```
+
+(change `python-driver` for the driver that you want to run). 
+
+Then you can send requests using any of the clients. Please note that bblfshd
+specific commands like `Version()` will not work in this mode.
```
abeaumont: AFAIK drivers support the same API as server, including version.

juanjux:
Yes, I tough the same, that page it's getting to unwieldy for a "Getting Started", I think I'm adding a new "Advanced Usage" section with this content and the "Adding Drivers from the local Docker daemon" one.

I'll also modify the [creating the integration tests](https://doc.bblf.sh/driver/sdk.html#creating-the-integration-tests) section to connect the `bblfsdh-sdk-tool` to a local driver instead of the server, since it's simpler and will give the same results as the test runner for things like empty files.

user/getting-started.md
juanjux: True! It works.

abeaumont:
sounds good.

Issue #111, opened 2017-11-08
Replace tmp by opt or a proper directory for macOS


abeaumont:

abeaumont:
Since `python-driver` doesn't work on macOS, we're going to indicate to install them inside the container instead.

PR #113, opened 2017-11-08
Change bblfshd mount point info

juanjux:

### DIFF
--- a/user/getting-started.md
+++ b/user/getting-started.md
@@ -34,9 +34,10 @@ add the `-v` parameter to keep the `/var/lib/bblfshd` directory in a volume:
 $ docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
-(On macOS remove the parameter `-v /var/lib/bblfshd:/var/lib/bblfshd` since the
+On macOS, remove the parameter `-v /var/lib/bblfshd:/var/lib/bblfshd` since the
 default case insensitive filesystem could conflict with the internal drivers'
-case sensitive one.)
+case sensitive one. In that case remember that since this directory will then be
+ephemeral, you'll need to reinstall the drivers after a server restart.
 
 If everything worked, `docker logs bblfshd` should output something like this:
 
--- a/user/getting-started.md
+++ b/user/getting-started.md
@@ -34,7 +34,9 @@ add the `-v` parameter to keep the `/var/lib/bblfshd` directory in a volume:
 $ docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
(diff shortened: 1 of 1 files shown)


user/getting-started.md
```diff
@@ -31,9 +31,11 @@ drivers will be lost when you stop the container. To avoid this from happening,
 add the `-v` parameter to keep the `/var/lib/bblfshd` directory in a volume:
 
 ```bash
-$ docker run -d --name bblfshd --privileged -p 9432:9432 -v /tmp/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
+$ docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
+(On macOS remove the parameter `-v /var/lib/bblfshd:/var/lib/bblfshd`.)
```
abeaumont: I think a bit more information here would be nice, to better understand why this is needed and how it will affect.

user/getting-started.md
```diff
@@ -31,9 +31,13 @@ drivers will be lost when you stop the container. To avoid this from happening,
 add the `-v` parameter to keep the `/var/lib/bblfshd` directory in a volume:
 
 ```bash
-$ docker run -d --name bblfshd --privileged -p 9432:9432 -v /tmp/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
+$ docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
+(On macOS remove the parameter `-v /var/lib/bblfshd:/var/lib/bblfshd` since the
+default case insensitive filesystem could conflict with the internal drivers'
+case sensitive one.)
```
abeaumont: I still miss some info about how this affects the user (no persistence). I don't think parenthesis are needed, either.

juanjux: Done!

abeaumont:
Fixes #111

user/getting-started.md
```diff
@@ -34,9 +34,10 @@ add the `-v` parameter to keep the `/var/lib/bblfshd` directory in a volume:
 $ docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
-(On macOS remove the parameter `-v /var/lib/bblfshd:/var/lib/bblfshd` since the
+On macOS, remove the parameter `-v /var/lib/bblfshd:/var/lib/bblfshd` since the
 default case insensitive filesystem could conflict with the internal drivers'
-case sensitive one.)
+case sensitive one. In that case remember that since this directory will then be
+ephemeral, you'll need to reinstall the drivers after a server restart.
```
abeaumont: It would be more precise to indicate _whenever a new container is created_, since you can restart the server without destroying the container, keeping the drivers.

juanjux: Restarting or creating the server container would be maybe more precise, since restarting a docker image (docker restart) also restarts it's ephemeral storage.

PR #121, opened 2017-11-28
use docker volumes to keep installed drivers

campoy:


juanjux:
Thanks for the PR! Could you try this with the Python driver? (it was the one that gave us problems with duplicated files with different casing).

juanjux:
CC @campoy  (I would test this myself but I'm not using a mac).

juanjux:
Tested, works.

smacker:
btw you don't really need `docker volume create bblfshd-cache`, docker will create it during `run`. But it's okay to have it in docs for better understanding what is going on.

PR #137, opened 2018-02-13
include Go driver

dennwc:
Note that docker image for golang is missing.


juanjux:
The container will be generated once we do a tagged release. I've already added the user/password settings for dockerhub in Travis for the go driver, so once you merge the PR and tag a release the CI for the release should upload to dockerhub.

PR #151, opened 2018-03-12
Request for comments for BIP-4 and BIP-5

dennwc:
Request for comments on:

- [x] **BIP-4:** Schema-less internal representation for AST
- [x] **BIP-5:** Shape-based AST transformations

These BIPs are sent together because, as we decided on the first design meeting, we will not wait for the first one to be accepted, and will implement both in one logical step.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -36,3 +36,5 @@
 * [BIP1: Purpose and Guidelines](proposals/bip-001.md)
 * [BIP2: Explicit Annotated status and completeness information](proposals/bip-002.md)
 * [BIP3: Agglutinative Roles language](proposals/bip-003.md)
+* [BIP4: Schema-less internal representation for AST](proposals/bip-004.md)
+* [BIP5: Shape-based AST transformations](proposals/bip-005.md)
--- a/proposals/README.md
+++ b/proposals/README.md
@@ -12,3 +12,5 @@ This is the index of Babelfish Improvement Proposals, known as BIPs.
 | 1      | Accepted | [BIP Purpose and Guidelines](bip-001.md)                             |
 | 2      | Accepted | [Explicit Annotated status and completeness information](bip-002.md) |
 | 3      | Accepted | [Agglutinative Roles language](bip-003.md)                           |
+| 4      | Accepted | [Schema-less internal representation for AST](bip-004.md)            |
+| 5      | Accepted | [Shape-based AST transformations](bip-005.md)                        |
--- a/proposals/bip-004.md
+++ b/proposals/bip-004.md
@@ -2,24 +2,24 @@
 
 | Field | Value |
 | --- | --- |
-| BIP | ??1 |
+| BIP | 4 |
 | Title | Schema-less internal representation for AST |
 | Author | Denys Smirnov |
-| Status | Draft |
+| Status | Accepted |
 | Created | 2018-02-23 |
-| Updated | 2018-02-23 |
-| Target version | 1.15 ?? |
+| Updated | 2018-03-14 |
+| Target version | 1.x |
 
 ## Abstract
 
 Current transformation code for AST/UAST nodes is based on Go structures
-that makes it harder to manipulate. Transformation code usually cannot be
+that make it harder to manipulate. Transformation code usually cannot be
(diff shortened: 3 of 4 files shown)


eiso:
Minor question, why maintain backwards compatibility for BIP 4 when BIP 5 introduces breaking changes and an SDK v2 anyways?

dennwc:
@eiso It was done before design meeting and I forgot to remove that part. It's one of the reasons why both proposal are here - if both will be accepted as-is, we can remove backward compatibility for BIP-4 and implement schema-less directly for BIP-5.

juanjux:
General comments:

There are some small grammar problems. Use grammarly, it's a great tool for we not natives.

Comments about BIP-5:

To get a fuller picture of the transformation and changed annotation DSLs, I would like to see a complete example, not in the sense of covering all the nodes and types of a language, but of seeing the integral source code of the transformations and annotations for two or three object types. Also, including some ASCII trees with the native AST and how it's transformed into the UAST would help (check [this](http://efanzh.org/tools/ascii-tree-graph-generator/) online tool for that). Then an explanation in prose about what are, step by step (high level steps, I mean) the changes done internally in the process.

Comments about BIP-4 coming soon.

juanjux:
**Comments on BIP-4:**

- In *As an example of uniform transformations, we will consider following transformations in ObjectToNode: InternalType and TokenKeys.* it would be clearer if it was specified that what you are explaining below that is how the current system works.

- In *Properties are saved directly into a parent object map.* I don't understand what is mean by *parent object map*. It's, for example, the same map where `@type` is stored? What happens if a property is called `@type` or `@token` too in the native AST?

- Same for the reference to the parent object map in Children. Also, what do you mean by *grouped by predicate*? Could you expand a little into that? (with maybe some simple example).

- Finally, I don't really see why there is a need for what is explained in this BIP (when trying to see it without relation to BIP-5) or what is the relation with the current priority of extracting higher level semantic objects. You "generalize away" the fields in the Node structure, but only for then saving them in the Object map with different names (also magic names, after all). Also, the internal type, token and positions are required by the current UAST specs (the position in a "relaxed" requirement) to be in the UAST nodes, so I don't see the problem with them being "hardcoded" properties of the `Node` structure. I'm not trying to be overly critic (sorry if I sound like that!), just asking you for helping me understand the need for this BIP.

dennwc:
@juanjux 

### BIP-4

1) Done

2) I'm referring to a way current implementation "flattens" object fields into a list. I want to revert this.
Currently, it works like this:
```
{name: foo, astType: ident}
->
type: ident
children:
  {irole: name, value: foo}
```
I propose to store all fields as-is, just rewrite the names:
```
{name: foo, @type: ident}
```
So answer for is that the map is the same where `@type` is stored, yes. It's highly unlikely that native AST will have such names since they usually are a serialization of native AST classes that can't have such names. But if this happens, we can provide a way to pre-process it and rewrite it to a different name.

3) Sorry for confusing it with annotation predicates. I refer to field names (aka internalRole) when I'm talking about predicates here.
Currently, it flattens array fields:
```
{name: foo, body: [{1}, {2}, {3}]}
->
children:
   {irole:name, val:foo}
   {irole:body, val:1}
   {irole:body, val:2}
   {irole:body, val:3}
```
And this structure is hard to manipulate.
I propose to revert this, and again, just store array in field as-is:
```
{name: foo, body: [{1}, {2}, {3}]}
->
{name: foo, body: [{1}, {2}, {3}]}
```
The only aspect that changes comparing to native AST is the type that is used to describe the tree (more strict then `map[string]interface{}`).

4) You are right about magical names and moving everything into Objects. But the reasoning behind this is that you cannot really implement transformations described in BIP-5 while having a strictly typed Node with few special-cased fields. Transformation code benefits from a generic approach because it can freely manipulate the structure without reflect package or hand-crafted code.

dennwc:
@juanjux Addressed comment for BIP-5

juanjux:
Awesome explanations, you resolved my doubts, thanks!

dennwc:
Will leave it open until we have a prototype

dennwc:
Rebased and squashed commits. Decided to merge it to be able to reference BIPs.

PR #154, opened 2018-03-14
Update a roles list generator and re-generate the list

### LINKED ISSUE #152: Update and auto-generate roles list
We should keep [./uast/roles.md](https://github.com/bblfsh/documentation/blob/master/uast/roles.md) file updated by auto-generating it from SDK sources and by examining available drivers.

dennwc:
Fixes #152

This can actually give a pretty good insight about driver completeness.

And it looks like there are roles that are not being used at all.

### DIFF
--- a/_tools/roles/main.go
+++ b/_tools/roles/main.go
@@ -5,13 +5,15 @@ import (
 	"bytes"
 	"fmt"
 	"go/ast"
-	"go/importer"
 	"go/parser"
 	"go/token"
-	"go/types"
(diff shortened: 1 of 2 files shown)


juanjux:
Don't worry about Unannotated and Incomplete not being used, that's actually a good thing ;).

I'm pretty sure Byte and Friend will have their uses, Friend is maybe too specific but if/when we do the two leveles Category/Roles is can remain.

Issue #35, opened 2017-06-06
Add simple comparison table for alternatives


bzz:
It would be nice to have a page where Babelfish is briefly compared with other similar efforts. Highlighting key differences would help people from other communities get interested in Babelfish as well as provide better understanding of possible solutions "landscape".

Other efforts worth comparing to:
 - [Microsoft/language-server-protocol](https://github.com/Microsoft/language-server-protocol) and [LSP Implementations](https://github.com/Microsoft/language-server-protocol/wiki/Protocol-Implementations)
 - [sourcegraph/srclib](https://github.com/sourcegraph/srclib)
 - [ctags](http://ctags.sourceforge.net/)
 - http://kythe.io

Something like https://github.com/OpenGrok/OpenGrok/wiki/Comparison-with-Similar-Tools or even simpler - a paragraph about each tool would be great. 

What would be the best place for such page? If you guys see value in this - I'll be happy to allocate some time and contribute a first draft.

smola:
Also, look at Hashicorp projects, which usually have a dedicated section to this:
https://www.consul.io/intro/vs/index.html
https://www.vagrantup.com/intro/vs/index.html

bzz:
Could be nice to include
 - https://github.com/antlr/antlr4
 - https://github.com/tree-sitter/tree-sitter

abeaumont:
Also:
* http://www.srcml.org/
* https://github.com/ThierryGoubier/SmaCC

dennwc:
Fixed in #155

PR #155, opened 2018-03-14
Describe how Babelfish compares to other projects

dennwc:
Draft. Addresses #35 to some degree. Still worth adding a comparison table before merging.

Please review carefully - my wording might not be correct. And I haven't tried all the systems described here, so the information might be wrong to some degree.

### DIFF
--- a/alternatives.md
+++ b/alternatives.md
@@ -127,9 +127,11 @@ From [project documentation](https://www.srcml.org/about.html):
 > The srcML format is an XML representation for source code, where the markup tags identify elements of the abstract syntax for the language. The srcml program is a command line application for the conversion source code to srcML, an interface for the exploration, analysis, and manipulation of source code in this form, and the conversion of srcML back to source code. Supports parsing C/C++, C#, and Java.
 
 srcML defines an [XML schema](https://www.srcml.org/doc/srcMLGrammar.html) to annotate
-source code files with it's syntactic structure: given the file with a source code, it
-converts it XML with that schema. It also provides tools written in C++ to [query](https://www.srcml.org/tutorials/xpath-query.html)
-and analyze files in this format using XPath. An AST structure is language-dependent.
+source code files with it's syntactic structure. Given a file with the source code, is
+converted to XML document, following that schema. srcML also provides tools in separate
+binaries written in C++ to [query](https://www.srcml.org/tutorials/xpath-query.html)
+and analyze files in this format using XPath. An AST structure is language-dependent
+and parsing itself is done using ANTLR.
 
 Babelfish provides a native language AST with positional information, that
 can be used to generate the same markup. Also, it provides a language-independent
--- a/alternatives.md
+++ b/alternatives.md
@@ -12,7 +12,7 @@ Babelfish and Kythe share a goal of defining a common representation
 for concepts from different programming languages and provide a way to query it.
 Both provide a unified data format. And both are language-independent.
 
-The main difference is that Babelfish preserve all AST nodes including
+The main difference is that Babelfish preserves all AST nodes including
 control flow and expressions, while Kythe focuses on class hierarchy,
 dependencies, etc.
 
@@ -53,7 +53,7 @@ structure will still allow querying for usages, etc.
 
 > Ctags generates an index (or tag) file of language objects found in source files that allows these items to be quickly and easily located by a text editor or other utility. A tag signifies a language object for which an index entry is available (or, alternatively, the index entry created for that object).
 
-Both Babelfish and Ctags provides positional information for identifiers,
+Both Babelfish and Ctags provide positional information for identifiers,
 classes, directives, etc.
 
 The main difference is that Ctags does not provide any form of AST, while
@@ -88,7 +88,7 @@ it preserves all features of original AST.
 
 ## [srcML](https://www.srcml.org/)
 
-> The srcML format is an XML representation for source code, where the markup tags identify elements of the abstract syntax for the language. The srcml program is a command line application for the conversion source code to srcML, an interface for the exploration, analysis, and manipulation of source code in this form, and the conversion of srcML back to source code. The current parsing technologies supports C/C++, C#, and Java.
+> The srcML format is an XML representation for source code, where the markup tags identify elements of the abstract syntax for the language. The srcml program is a command line application for the conversion source code to srcML, an interface for the exploration, analysis, and manipulation of source code in this form, and the conversion of srcML back to source code. The current parsing technologies support C/C++, C#, and Java.
 
 srcML defines an XML schema to annotate source code files with AST structure.
 It also provides tools to query and analyze files in this format. An AST
--- a/alternatives.md
+++ b/alternatives.md
@@ -1,8 +1,20 @@
 # Babelfish vs Other Software
 
+Table of Contents
+=================
+
+* [Kythe](#kythe)
+* [Language Server Protocol](#language-server-protocol)
+* [srclib](#srclib)
+* [ctags](#ctags)
+* [ANTLR](#antlr)
+* [Tree\-sitter](#tree-sitter)
+* [srcML](#srcml)
+* [SmaCC](#smacc)
+
 <!-- TODO: https://github.com/oracle/opengrok/wiki/Comparison-with-Similar-Tools -->
 
-## [Kythe](https://kythe.io/)
+## [Kythe](http://kythe.io/)
 
 > The best way to view Kythe is as a “hub” for connecting tools for various languages, clients and build systems. By defining language-agnostic protocols and data formats for representing, accessing and querying source code information as data, Kythe allows language analysis and indexing to be run as services.
 
@@ -16,15 +28,19 @@ The main difference is that Babelfish preserves all AST nodes including
 control flow and expressions, while Kythe focuses on class hierarchy,
 dependencies, etc.
 
-Also, Babelfish provides a unified IR across languages, which is defined
+Babelfish provides a unified IR across languages, which is defined
 as a non-goal of Kythe.
 
-Kythe requires to instrument language compilers and build systems while
-Babelfish uses native language parsers to get an AST, allowing to develop
-a new language driver in less time.
+At the same time since Babelfish only processes one file at the time, it cannot
+resolve symbolic references in the code, while the goal on Kythe project is to
+provide this information with a very high accuracy.
 
-Kythe also provides a way to process the whole project, while for now
-Babelfish is focused on processing individual files.
+Kythe requires to instrument language compilers and build systems (by modifying
+the sources of compiler and build system), while Babelfish uses native language
+parsers to get an AST, allowing to develop a new language driver in less time.
+
+Kythe requires a project to be built to be processed because of integration with
+compilers and build systems, while Babelfish can process only individual files.
 
 ## [Language Server Protocol](https://microsoft.github.io/language-server-protocol/)
 
@@ -37,9 +53,9 @@ Usages, etc. But it does not define a common representation of AST because
 the goal of the project is to enable easy access to analysis that is done
 by compilers and language SDKs.
 
-Babelfish provides a common representation for ASTs allowing
-to use its output for static analysis by other tools. Queries over UAST
-structure will still allow querying for usages, etc.
+Babelfish provides a common representation for ASTs allowing to use its output
+for static analysis by other tools. Babelfish cannot resolve symbolic references,
+thus it cannot be used to query usages.
 
 ## [srclib](https://srclib.org/)
 
--- a/alternatives.md
+++ b/alternatives.md
@@ -16,6 +16,7 @@ Table of Contents
 
 ## [Kythe](http://kythe.io/)
 
+From official [Kythe documentation](https://kythe.io/docs/kythe-overview.html#_goals_of_kythe):
 > The best way to view Kythe is as a “hub” for connecting tools for various languages, clients and build systems. By defining language-agnostic protocols and data formats for representing, accessing and querying source code information as data, Kythe allows language analysis and indexing to be run as services.
 
 > Some tools (e.g., static analyzers) already have expressive purpose-built internal representations for code. Kythe is not meant to be a universal replacement for such IRs — instead, our goal is to provide a way for such tools to capture “interesting subsets” of an analysis for sharing with other tools.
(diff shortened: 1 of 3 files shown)


alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+
+> The best way to view Kythe is as a “hub” for connecting tools for various languages, clients and build systems. By defining language-agnostic protocols and data formats for representing, accessing and querying source code information as data, Kythe allows language analysis and indexing to be run as services.
+
+> Some tools (e.g., static analyzers) already have expressive purpose-built internal representations for code. Kythe is not meant to be a universal replacement for such IRs — instead, our goal is to provide a way for such tools to capture “interesting subsets” of an analysis for sharing with other tools.
+
+Babelfish and Kythe share a goal of defining a common representation
+for concepts from different programming languages and provide a way to query it.
+Both provide a unified data format. And both are language-independent.
+
+The main difference is that Babelfish preserve all AST nodes including
```
juanjux: preserves

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+
+> srclib handles: package detection, global dependency resolution, type inference, querying the graph of definitions and references in code, versioning using different VCS systems, and semantic blaming.
+
+<!-- TODO -->
+
+## [ctags](http://ctags.sourceforge.net/)
+
+> Ctags generates an index (or tag) file of language objects found in source files that allows these items to be quickly and easily located by a text editor or other utility. A tag signifies a language object for which an index entry is available (or, alternatively, the index entry created for that object).
+
+Both Babelfish and Ctags provides positional information for identifiers,
```
juanjux: provide

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+analysis. Instead, it focuses on performance and implements real-time
+tree diffing.
+
+Babelfish provides a full native language AST as well as language-independent
+UAST for all supported languages. It is suitable for static analysis since
+it preserves all features of original AST.
+
+## [srcML](https://www.srcml.org/)
+
+> The srcML format is an XML representation for source code, where the markup tags identify elements of the abstract syntax for the language. The srcml program is a command line application for the conversion source code to srcML, an interface for the exploration, analysis, and manipulation of source code in this form, and the conversion of srcML back to source code. The current parsing technologies supports C/C++, C#, and Java.
```
juanjux: support

dennwc: The fun thing is that it's the quote from their site.

dennwc:
@juanjux Done

juanjux:
I'll wait for @eiso review to merge since he has a lot of knowledge about Kythe and its people.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+dependencies, etc.
+
+Also, Babelfish provides a unified IR across languages, which is defined
+as a non-goal of Kythe.
+
+Kythe requires to instrument language compilers and build systems while
+Babelfish uses native language parsers to get an AST, allowing to develop
+a new language driver in less time.
+
+Kythe also provides a way to process the whole project, while for now
```
bzz: May be also worth mentioning here that Kythe depends on ability to build the project, while bblfsh does not require build at all.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+Babelfish is focused on processing individual files.
+
+## [Language Server Protocol](https://microsoft.github.io/language-server-protocol/)
+
+> A Language Server is meant to provide the language-specific smarts and communicate with development tools over a protocol that enables inter-process communication.
+
+> The idea behind the Language Server Protocol (LSP) is to standardize the protocol for how such servers and development tools communicate. This way, a single Language Server can be re-used in multiple development tools, which in turn can support multiple languages with minimal effort.
+
+LSP defines a common protocol and RPC for queries like Go-To-Definition,
```
bzz: 👍 May be also worth mentioning that it's a protocol and and API exposed though client libraries from https://langserver.org/

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+to use its output for static analysis by other tools. Queries over UAST
+structure will still allow querying for usages, etc.
+
+## [srclib](https://srclib.org/)
+
+> srclib makes developer tools like code search and static analyzers better. It supports things like jump to definition, find usages, type inference, and documentation generation.
+
+> srclib handles: package detection, global dependency resolution, type inference, querying the graph of definitions and references in code, versioning using different VCS systems, and semantic blaming.
+
+<!-- TODO -->
```
bzz: AFAIK this project is not maintained by sourcegraph any more.

But I think it might worth spend a little bit of time here to highlight the difference, as one can see other companies in source code analysis field are using it i.e https://fossa.io/ though https://github.com/fossas/srclib and it would be nice to understand why bblfsh would not feat for them.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+
+> ANTLR (ANother Tool for Language Recognition) is a powerful parser generator for reading, processing, executing, or translating structured text or binary files. It's widely used to build languages, tools, and frameworks. From a grammar, ANTLR generates a parser that can build parse trees and also generates a listener interface (or visitor) that makes it easy to respond to the recognition of phrases of interest.
+
+ANTLR is a language generator with support for most existing languages
+and data formats. But parsers usually produce parse trees that are usually
+different from native language AST, thus requiring additional processing
+to be used for analysis. The structure of parse tree is also language-dependent.
+
+Babelfish provides a correct native language AST as well as
+language-independent UAST for all supported languages.
```
bzz: May be worth mentioning here that bblfsh architecture would allow to build native drivers that use ANTLR grammars instead, but on top of difference \w native language AST, that would have the burden of supporting the grammar (or something like that?).

This would highlight (and may be link to?) nice aspects of bblfsh architecture.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+
+## [Tree-sitter](https://github.com/tree-sitter/tree-sitter)
+
+> Tree-sitter is a C library for incremental parsing, intended to be used via bindings to higher-level languages. It can be used to build a concrete syntax tree for a program and efficiently update the syntax tree as the program is edited. This makes it suitable for use in text-editing programs.
+
+Tree-sitter provides a simplified AST to be able to execute a limited set
+of queries. It usually cannot provide enough details suitable for static
+analysis. Instead, it focuses on performance and implements real-time
+tree diffing.
```
bzz: @maxbrunsfeld would know better, but it seems that tree-sitter is an awesome project that is designed to be used in editors, so is optimized for the incremental parsing and low latencies, both of which are not the problems we have in large-scale parsing scenario that bblfsh is addressing.

Tree-sitter also requires maintaining custom grammar definitions for each supported language, that bblfsh does not.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+> The srcML format is an XML representation for source code, where the markup tags identify elements of the abstract syntax for the language. The srcml program is a command line application for the conversion source code to srcML, an interface for the exploration, analysis, and manipulation of source code in this form, and the conversion of srcML back to source code. The current parsing technologies support C/C++, C#, and Java.
+
+srcML defines an XML schema to annotate source code files with AST structure.
+It also provides tools to query and analyze files in this format. An AST
+structure is language-dependent.
+
+Babelfish provides a native language AST with positional information, that
+can be used to generate the same markup. Also, it provides a language-independent
+UAST for all supported languages. It allows performing the same queries
+and analysis for different programming languages.
```
bzz: Sorry, the distinction \w srcML is not super-clear to me from this summary, except for different approach with XML schema for annotations :/

bzz:
This list looks great, thank you @dennwc for stepping up and preparing it! 🚀 

Do you think we should add a ToC at the beginning?  In past I used this [CLI MD ToC generator](https://github.com/ekalinin/github-markdown-toc#local-files) it worked quite well.

Few minor comments above. Also CI seems to be un-happy about some external links <details>

```
$ docker run -t -v $(pwd)/_book:/site 18fgsa/html-proofer /site
Unable to find image '18fgsa/html-proofer:latest' locally
latest: Pulling from 18fgsa/html-proofer
Status: Downloaded newer image for 18fgsa/html-proofer:latest
Running ["ImageCheck", "ScriptCheck", "LinkCheck"] on ["/site"] on *.html... 
Checking 111 external links...
Ran on 22 files!
- /site/alternatives.html
  *  External link https://kythe.io/ failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: Peer certificate cannot be authenticated with given CA certificates
htmlproofer 3.8.0 | Error:  HTML-Proofer found 1 failure!
The command "docker run -t -v $(pwd)/_book:/site 18fgsa/html-proofer /site" exited with 1.
```
</details>

<br />
\cc and adding to reviewers @smola who one of the original authors of bblfsh.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+Also, Babelfish provides a unified IR across languages, which is defined
+as a non-goal of Kythe.
+
+Kythe requires to instrument language compilers and build systems while
+Babelfish uses native language parsers to get an AST, allowing to develop
+a new language driver in less time.
+
+Kythe also provides a way to process the whole project, while for now
+Babelfish is focused on processing individual files.
+
```
eiso: I would add here that one of Kythe's main goals is to provide symbolic cross-referencing with as high accuracy as possible. Which is part of their mission to be the common format on which all developer tooling could be built. This is what leads them to need to process input from compilers and build systems. Babelfish today does not share that same goal, while in the future we might want to resolve references, but likely not to the extent that Kythe aims to do so. 

I would also mention that Kythe has worked with authors of compilers and build systems to be able to get the necessary information while Babelfish doesn't.

alternatives.md
```diff
@@ -0,0 +1,113 @@
...
+LSP defines a common protocol and RPC for queries like Go-To-Definition,
+Usages, etc. But it does not define a common representation of AST because
+the goal of the project is to enable easy access to analysis that is done
+by compilers and language SDKs.
+
+Babelfish provides a common representation for ASTs allowing
+to use its output for static analysis by other tools. Queries over UAST
+structure will still allow querying for usages, etc.
+
```
eiso: A similar point as above, Babelfish today does not care about resolving objects/cross-references.

alternatives.md
eiso: AFAIK the main reseason FOSSA uses srclib and would not be able to use Babelfish is the dependency graph it can generate through symbolic cross-references (sorry for being a one trick pony answering this everywhere but it seems relevant).

alternatives.md
eiso: +1 on @bzz comment. Tree-Sitter is a very impressive effort but has the downside of creating/maintaining grammars. For its use case, it's perfect. Babelfish is not meant to be used at editor level. For this, I would always recommend tree-sitter.

dennwc: I did not mean to discourage the use of tree-sitter project, sorry if it sounds like this. The project is indeed impressive.

I will update the description to mention custom grammars and all the good features it has.

At the same time, I tried running it, and it seems like an AST that is generated is not a real AST of the language. I wouldn't say it's just a parse tree like in ANTLR, but it missing some details that native AST has.

alternatives.md
dennwc: Will try to clarify it. Need some time to check features of each system.

alternatives.md
dennwc: Okay, will drop all future references from this document

dennwc:
@eiso Sorry, it was not meant to sell bblfsh as a silver bullet but was assuming a particular use case.

I will add clarification on what use case is covered by each system.

alternatives.md
dennwc: Will mention build systems explicitly - it was assumed.

@eiso Also it would be nice to hear from you about the meeting with Kythe developers to know more about the goal of the project in more details.

dennwc:
@bzz Thanks, I will include a ToC. The document indeed might get quite long.

The link is actually correct, but Travis cannot recognize root cert for some reason. I will switch the link to http scheme.

alternatives.md
mcuadros: Isn't maintained. The abandoned in favor of LSP, AFAIK

alternatives.md
mcuadros: Also the generated parser/lexer are language dependant. AFAIK

alternatives.md
mcuadros: Also is oriented to support partial/malformed code, since is made for editors.

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+* [ANTLR](#antlr)
+* [Tree\-sitter](#tree-sitter)
+* [srcML](#srcml)
+* [SmaCC](#smacc)
+
+<!-- TODO: https://github.com/oracle/opengrok/wiki/Comparison-with-Similar-Tools -->
+
+## [Kythe](http://kythe.io/)
+
+> The best way to view Kythe is as a “hub” for connecting tools for various languages, clients and build systems. By defining language-agnostic protocols and data formats for representing, accessing and querying source code information as data, Kythe allows language analysis and indexing to be run as services.
```
campoy: Is this quote directly from the documentation?
Could you say it in the doc and add a link?

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+
+> Some tools (e.g., static analyzers) already have expressive purpose-built internal representations for code. Kythe is not meant to be a universal replacement for such IRs — instead, our goal is to provide a way for such tools to capture “interesting subsets” of an analysis for sharing with other tools.
+
+Babelfish and Kythe share a goal of defining a common representation
+for concepts from different programming languages and provide a way to query it.
+Both provide a unified data format. And both are language-independent.
+
+The main difference is that Babelfish preserves all AST nodes including
+control flow and expressions, while Kythe focuses on class hierarchy,
+dependencies, etc.
```
campoy: Rather than class hierarchy I'd say it focuses on object resolution, linking usages of identifiers to their original definition.

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+Kythe requires to instrument language compilers and build systems (by modifying
+the sources of compiler and build system), while Babelfish uses native language
+parsers to get an AST, allowing to develop a new language driver in less time.
+
+Kythe requires a project to be built to be processed because of integration with
+compilers and build systems, while Babelfish can process only individual files.
+
+## [Language Server Protocol](https://microsoft.github.io/language-server-protocol/)
+
+> A Language Server is meant to provide the language-specific smarts and communicate with development tools over a protocol that enables inter-process communication.
```
campoy: Same, link to docs

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+> The idea behind the Language Server Protocol (LSP) is to standardize the protocol for how such servers and development tools communicate. This way, a single Language Server can be re-used in multiple development tools, which in turn can support multiple languages with minimal effort.
+
+LSP defines a common protocol and RPC for queries like Go-To-Definition,
+Usages, etc. But it does not define a common representation of AST because
+the goal of the project is to enable easy access to analysis that is done
+by compilers and language SDKs.
+
+Babelfish provides a common representation for ASTs allowing to use its output
+for static analysis by other tools. Babelfish cannot resolve symbolic references,
+thus it cannot be used to query usages.
```
campoy: "query usages" is ambiguous here, maybe "query the usages of a given identifier"?

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+Babelfish provides native language AST as well as language-independent UAST.
+
+Also, Babelfish ecosystem allows making more complex queries over AST.
+
+## [ANTLR](https://github.com/antlr/antlr4)
+
+> ANTLR (ANother Tool for Language Recognition) is a powerful parser generator for reading, processing, executing, or translating structured text or binary files. It's widely used to build languages, tools, and frameworks. From a grammar, ANTLR generates a parser that can build parse trees and also generates a listener interface (or visitor) that makes it easy to respond to the recognition of phrases of interest.
+
+ANTLR is a language generator with support for most existing languages
+and data formats. But parsers usually produce parse trees that are usually
```
campoy: drop one of the "usually"

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+different from native language AST, thus requiring additional processing
+to be used for analysis. The structure of parse tree is also language-dependent.
```
campoy: `of the parse tree`, or `of parse trees`

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+analysis. Instead, it focuses on performance and implements real-time
+tree diffing.
+
+Babelfish provides a full native language AST as well as language-independent
+UAST for all supported languages. It is suitable for static analysis since
+it preserves all features of original AST.
+
+## [srcML](https://www.srcml.org/)
+
+> The srcML format is an XML representation for source code, where the markup tags identify elements of the abstract syntax for the language. The srcml program is a command line application for the conversion source code to srcML, an interface for the exploration, analysis, and manipulation of source code in this form, and the conversion of srcML back to source code. The current parsing technologies support C/C++, C#, and Java.
```
campoy: srcML appears also as srcml, is that intended?

dennwc:
@campoy Thanks for all the comments! I will address them as soon as I finish few LA-related tasks.

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+
+Both Babelfish and Ctags provide positional information for identifiers,
+classes, directives, etc.
+
+The main difference is that Ctags does not provide any form of AST, while
+Babelfish provides native language AST as well as language-independent UAST.
+
+Also, Babelfish ecosystem allows making more complex queries over AST.
+
+## [ANTLR](https://github.com/antlr/antlr4)
```
smola: Should this section be about ANTLR and other parser generators? Or include a mention about this being applicable to comparison with other parser generators?

alternatives.md
```diff
@@ -0,0 +1,129 @@
...
+srcML defines an XML schema to annotate source code files with AST structure.
+It also provides tools to query and analyze files in this format. An AST
+structure is language-dependent.
+
+Babelfish provides a native language AST with positional information, that
+can be used to generate the same markup. Also, it provides a language-independent
+UAST for all supported languages. It allows performing the same queries
+and analysis for different programming languages.
+
+## [SmaCC](http://www.refactoryworkers.com/SmaCC.html)
```
smola: Same as with ANTLR

campoy:
Hey @dennwc let me know when you're ready for more reviews

juanjux:
This would be a worthy addition to the documentation, @dennwc what is the current state of this PR?

dennwc:
@juanjux Still need to address some review comment and rewrite few sections. Will do it after we finish porting drivers.

bzz:
🙏 as this great work is almost ready!

dennwc:
We are almost done with semantic drivers, so I will revisit it soon. Probably even next week.

alternatives.md
maxbrunsfeld: > Tree-sitter provides a simplified AST to be able to execute a limited set of queries.

You're right that the output is not an AST, but it's not simplified; it's a Concrete Syntax Tree, so it's a superset of the information contained in an AST. ANTLR produces a CST too, but I don't think ANTLR's trees are designed to use directly for code analysis, whereas Tree-sitter's actually are.

The goal with Tree-sitter grammars is to closely mirror the structure of the AST, so that the resulting tree is *similar* to an AST, but with additional *anonymous* tokens present as well, a little bit like [the ESTree](https://github.com/estree/estree) format for JavaScript syntax trees. [Tree-sitter's APIs](http://tree-sitter.github.io/tree-sitter/using-parsers#named-vs-anonymous-nodes) let you ignore the anonymous tokens if you want.

That said, there is still a big difference which may be important for your use case: Tree-sitter's syntax nodes currently just contain an ordered list of children. The syntax nodes don't have a name for each child like `predicate`, `body`, etc.

dennwc: @maxbrunsfeld Thanks for the clarification! I will update the document accordingly.

campoy:
Hey @dennwc, it's maybe a good time to revisit?

bzz:
@campoy as discussed IRL \w @dennwc I will be picking up this great work and trying to finish it ASAP.
Stay tuned!

dennwc:
@bzz Thank you for taking this piece of work :) Unfortunately I have no time to finish it right now.

alternatives.md
bzz: Clarified explicit XML-conversion machinery as the biggest difference.

alternatives.md
bzz: 👍 Added a note about Kythe's main use case that drives the difference.

juanjux:
Great! I'll review it soon, but note that there are conflicts on two files.

bzz:
> We should probably ask Michael to review and improve the kythe section once he joins.
Indeed, great idea, we will! 

Also mind that some further improvements are expected soon e.g adding https://github.com/SonarSource/slang#slang as soon as these changes lands first.

@campoy could please take another look?

bzz:
Rebased on the latest master

bzz:
🚀 https://doc.bblf.sh/alternatives.html

PR #156, opened 2018-04-13
Update docs for BIP-5

dennwc:
* update BIP-5 description to more closely match final implementation (naming and syntax details)

* add a guide for transformation DSL

Work in progress.


driver/annotations.md
```diff
@@ -1,17 +1,23 @@
 # Adding UAST Annotations
```
juanjux: I think that the old annotation DSL should be removed from this file and probably be replaced (in this same file) by what you wrote on the transformation DSL file.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+```
+
+For the "name" field the state is required to reconstruct it:
+
+```golang
+"name": Var("x"),
+```
+
+At this point, state variable "x" will be used to reconstruct the field
+value (assign it to null).
```
juanjux: maybe "and assign null to it" or "and set it to null".

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+a string, an array or another AST node.
+
+```json
+{
+    "type": "Ident",
+    "name": null
+}
+```
+
+Since there is no more constraints defined, `Part` transform will use
```
juanjux: **the** `Part` transform...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+it's state stored in "other" variable to populate all unhandled fields:
```
juanjux: in **the ** "other" variable...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+{
+    "type": "Ident",
+    "name": null,
+    "offset": 5
+}
+```
+
+As you can see, an operation can be used to describe a tree shape and
+reconstruct it back, assuming that all placeholder variables used in
+an operation are defined in node state.
```
juanjux: in **the** node state || in node's state.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+s := NewState()
+
+// verify an AST with operation A, populate state
+if ok, _ := A.Check(s, ast); ok {
+    // construct a new AST using the state and operation B
+    newAST, _ := B.Construct(s, nil)
+}
+```
+
+As long as all variable defined in A exactly match all variables defined
```
juanjux: "as all variable**s**..."

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+in B, `Check`-`Construct` routine will work the same way as described
```
juanjux: **the** `Check` - `Construct` routine will work **in** the same way as described above (or in **the** example section above).

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+in example section above.
+
+Since state only stores placeholder values, while operation defines the
```
juanjux: Since **the** state... while **the** operation...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+shape of the tree, any complex shape transformation can be defined:
+
+```golang
```
juanjux: Not review, just wanted to say that this example showing the code and the json nodes is brilliant and shows pretty clearly how the transformations and checks operate in a very simple way.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+    },
+    "left": {"type":"Ident", "name": "v"},
+    "right": 5
+}
+```
+
+
+## Restrictions
+
+DSL also enforces few restrictions on transformations and operations.
```
juanjux: **The** DSL also enfoces **a** (or some) few restrictions...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+### Completeness
+
+All operations should either:
+
+* Reject the tree if it does not match one of constraints.
+
+* Describe the whole tree, and save unknown/placeholder nodes to the state.
+
+This rule ensures that if an operation matches the tree, it would be able
+to replicate it from a state associated with the node, since all fields
```
juanjux: Missing comma after "that" and spurious comma after "node".

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+are covered by it.
+
+This property allows to test each transformation by executing
+`Check`-`Construct` routine, or detect if a mapping defined by a driver
```
juanjux: **the** `Check`-`Construct` routine... by a driver**'s** developer... (**which** would result in an error).

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+developer ignores certain AST fields that are seen in the wild
+(will result in an error).
+
+### Reversibility
+
+Since all operations are required to implement both `Check` and `Construct`
+they can be used interchangeably as either source and/or target transforms.
+
+In simple case it means that transform A can produce the state S for AST
```
juanjux: I would rephrase as:

"For a simple case, this means, that given a node T over which we apply a transform A, we can produce a state S (with `Check`) and then reconstruct the original node T (with `Construct`) by only using the state S."

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+node T (with `Check`), and then reconstruct an original node T (with `Construct`)
+by only using the state S.
+
+```
+S := A.Check(T); A.Construct(S) == T
+```
+
+In a more complete example, it's possible for transform A to check
```
juanjux: for **the** transform A (specially important in this case since without the the transform could be read as a verb).

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+the tree T, produce the state S and for transform B to use this state
```
juanjux: **which** produces...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+T == A.Construct(S')
+```
+
+## List of operations
+
+### Values
+
+#### Any value
+
+In case any value can match the transformation, it should be stored into
```
juanjux: "If any value can...".

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+a named variable: `Var("x")`. Any name can be used for the variable as
+long as it is unique in this transformation.
+
+When writing mappings, variable with the same name should exist in the
```
juanjux: **a** variable with the same name

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+second transformation.
+
+#### Equal
+
+To check value for exact match, `Is(v)` operation is used.
+
+SDK defines helpers for common types: `String(v)`, `Int(v)`, etc.
```
juanjux: **The** SDK

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+#### Equal or nil
+
+To check any node for nil or specific rule `Opt` is used:
+
+```
+Opt("is_nil", Is(42))
+```
+
+`Opt` needs a named variable to store it's state ("is_nil" in this case).
```
juanjux: Mmmm this could use a small example. Also, Opt can be used not only for nil values but also for values that can be missing, right? If that's the case that should be written too.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+#### Value in
+
+To check that value equal any value in the list, two operations should be
```
juanjux: that **a** value equal**s** (unsure if that's a verb, if not "is equal to").

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+executed:
+
+1) Check that value is in the list.
```
juanjux: that **the** value...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+2) Store the value into a variable.
+
+To express this, `Check` helper can be used:
```
juanjux: **the** `Check` helper...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+```golang
+Check(
+    // condition
+    In(
```
juanjux: Are the conditions inside the `In` evaluated as an "and" or "or"? In any case, that could be specified (my guess is that it's an "or" since the func is called "In" but being a little explicit won't hurt in this case).

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+Check(
+    // condition
+    Not(String("A")),
+
+    // store the value
+    Var("x"),
+)
+```
+
+Note that value can still be of any type and type constraint of `String`
```
juanjux: that **the** value... also the part after and "and" is  a little messy, maybe:

 "Note that the value of any type since we're using `Var` and thus the type constraint of `String` inside the `Check` will not affect the stored variable".

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+will not affect the variable.
+
+The same way a value can be checked for "not in" constraint:
```
juanjux: **In** the same way**,**... for **a/the**...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+```
+Opt("is_nil", Obj{ ... })
+```
+
+`Opt` needs a named variable to store it's state ("is_nil" in this case).
+
+#### All fields must exist
+
+For object that always have a specific set of fields, `Obj` operation
```
juanjux: **the** `Obj`...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+can be used:
+
+```golang
+Obj{
+    "type": String("Ident"),
+    "name": Var("name"),
+}
+```
+
+Since this operation is represented by a map, order of operations is not
```
juanjux: **the** order...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+preserved. Operations will be executed in order of field name sorting.
```
juanjux: "will be executed following an alphabetical order from the field name"

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+This might affect operations that requres a variable to be defined, such
```
juanjux: requres => requires

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+as `Lookup` and `If`. To control execution order, use `Fields`.
```
juanjux: **the** execution order

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+#### Specific field options
+
+To get more control of how object fields are checked, use `Fields`:
```
juanjux: maybe: "about how **the** ordering and other aspects of the fields' checks..."

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+```golang
+Fields{
+    {Name: "type", Op: String("If")},
+    {Name: "then": Op: Var("then")},
+    {Name: "else": Op: Var("then"), Optional: "else_exists"},
+}
+```
+
+For each optional field a variable should be provided to store the state.
```
juanjux: Comma after field.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+Optional field means that they may not exist. For checking nil fields see
+`Opt` operation.
+
```
juanjux: Since Field is pretty important, could use one of those expanded examples with the json representations of the objects before and after the transformation.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+#### Skip unknown fields
+
+To skip unknown fields in object `Part` is used:
```
juanjux: I would rephrase as: `Part` can be used to skip unknown fields in an object:

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+```golang
+Part("other", Obj{
+    "name": Var("x"),
+})
+```
+
+`Part` will store all unhandled fields in `other` variable. It usually
```
juanjux: in **the** `other` variable. It usually required to have **a** `Part` call on both sides of a mapping (also, is this "usually" or always?).

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+requires to have `Part` on both sides of a mapping.
+
+### Arrays
+
+#### Fixed number of elements
+
+To represet an array with a specific number of elements, `Arr` can be used:
```
juanjux: Sounds more natural with the "`Arr` can be used..." at the start. Also, typo on represent.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+```golang
+Arr(
+    // element 0, fixed
```
juanjux: Maybe expand a little with "// element 0, fixed, must be an int with a value of 42".

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+```golang
+Each("elems",
+    Obj{
+        "name": Var("x"),
+    },
+)
+```
+
+Note that `Each` requires a named variable to store elements. This also
```
juanjux: **the** elements.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+means that `Each` should always be on both side of a mapping.
```
juanjux: side => sides.

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+
+`Each` allows an array to be empty or nil.
+
+Operation that is passed to `Each` will receive a clean state, thus
```
juanjux: **The** Operation that...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+it will not be able to access any variables from the parent. It also
+means that variable names inside this operation will not collide with
+the variables defined by parent transformation.
+
```
juanjux: by **the** parent...

driver/transform-dsl.md
```diff
@@ -0,0 +1,549 @@
...
+)
+```
+
+`Append` will verify that node is an array and will filter out "suffix"
+nodes before passing it to primary operation (`Var("x")` in this case).
+
+When used to costruct the tree, it will append "suffix" nodes to
+the end of an array defined by sub-operation.
+
+As in Go, `Append` will not affect variable "x".
```
juanjux: Not about this specific line: missing some doc and examples for `LookupArrOpVal` and friends.

proposals/bip-005.md
```diff
@@ -88,35 +87,29 @@ to reason about the transformation in terms of writing nodes, as if
 a function call is not describing the structure, but constructing it:
 get variable `off`, convert to string, store it as the key `OffsetKey`.
 
-At the end, you can see a `OtherKeys` helper that will save all other
-node keys that were not used in previous transformations.
+You can see a `Part` helper wraps the main transformation. It will save
```
juanjux: You can see **that**...

proposals/bip-005.md
```diff
@@ -128,24 +121,20 @@ Note that this time str-to-int conversion is omitted - it only happens
 on one side of the transformation, because variable should store a
```
juanjux: because **the/a** variable...

proposals/bip-005.md
```diff
@@ -128,24 +121,20 @@ Note that this time str-to-int conversion is omitted - it only happens
...
 "canonical" representation of that value, as defined by final tree shape.
 
-Both tree shapes are provided as an argument to `Transformer` that will
-bind transformations together:
+Both tree shapes are provided as an argument to `Map` that will
```
juanjux: Maybe:  as an argument => as arguments

proposals/bip-005.md
```diff
@@ -128,24 +121,20 @@ Note that this time str-to-int conversion is omitted - it only happens
...
+bind transformations together creating a mapping:
```
juanjux: bind **the**...

proposals/bip-005.md
```diff
@@ -247,24 +235,21 @@ conversion, etc.
 * `Check(sel []Sel, op Op)` - checks all sel arguments before executing
   op's check. Reversal only executes apply of op. Mostly used for `Not`.
 
-* `Object(op Op)` - verifies that current node is an object and checks
-  it with provided op. Reversal changes node type to object and applies
-  a provided operation to it. This operation will populate a list of
+* `Obj map[string]Op` - verifies that current node is an object and checks
```
juanjux: that **the** current...

dennwc:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 75 unresolved discussions.

---

*[driver/annotations.md, line 1 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mivF1Vy3MnDUmr-r1-1:-LANStJB8OLX4bNzKKZa:b-5tyx40) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/annotations.md#L1)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

I think that the old annotation DSL should be removed from this file and probably be replaced (in this same file) by what you wrote on the transformation DSL file.
</blockquote></details>

Yes, the plan was to update that section with a guide for new helpers as well. Still WIP.

---

*[driver/transform-dsl.md, line 29 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-29:-LANQarK4jrE2IfbphV9:b-hlgf0y) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L29)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

This could be removed and the comments added to the snippet above.
</blockquote></details>

It was actually intentional :) Wanted to focus reader's attention one detail at a time - it might be easier to understand. But can revert it if you think it's okay in one snippet.

---

*[driver/transform-dsl.md, line 106 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-106:-LANWkJdDnah7Eq8SZfK:b-70p2n3) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L106)):*
> probably should also be specified that String does a check but also stores that given string in the shared state.

It's not :) Because this state is "static" (baked into this operation), thus it doesn't need to be stored.

---

*[driver/transform-dsl.md, line 136 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-136:-LANRMR4Bqf6PKqdAflk:bro00ib) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L136)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Part, if used, would replace `Map`, true? In that case this could be specified (something like "as partial, using a `Part` call instead of `Map` which...").
</blockquote></details>

No, it still requires a `Map`. Should I write the full snippet with `Map` to make it more clear?

---

*[driver/transform-dsl.md, line 374 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-374:-LANahCGE62_Yjq5GI6h:b-kb1p64) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L374)):*
> Also, Opt can be used not only for nil values but also for values that can be missing, right?

If by "missing values" you mean "missing fields", then no. There was a note above:

> Note that the DSL makes a distinction between null fields and fields that does not exist.

Now I added few more details. Is it more clear this way?

---

*[driver/transform-dsl.md, line 390 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-390:-LANc4oM1StjK62H8jou:bkas5sv) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L390)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Are the conditions inside the `In` evaluated as an "and" or "or"? In any case, that could be specified (my guess is that it's an "or" since the func is called "In" but being a little explicit won't hurt in this case).
</blockquote></details>

`In` is a specific helper that uses "or" logic, as a user would expect from a "value in ..." operation.
The `Check` itself has only one condition, so if it would be an `And`/`Or`, it would be mentioned explicitly.

---

*[driver/transform-dsl.md, line 489 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-489:-LANeCbLDXaebM0UHcJR:be4ia00) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L489)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

in **the** `other` variable. It usually required to have **a** `Part` call on both sides of a mapping (also, is this "usually" or always?).
</blockquote></details>

Usually. Because you can have something like this:

```golang
Part("other", Obj{
  "name": Var("name"),
})
// ->
Obj{
  "name": Var("name"),
  "unrecognized": Var("other"),
}
```

Although I don't want to mention it because it's a bit more like implementation detail right now. If we will have a driver that relies on it, I will definitely describe it (and make it an API contract for `Part`).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156)*

juanjux:
*[driver/transform-dsl.md, line 29 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-29:-LAS7s3ZCqMEtwjUNFuc:b-dod832) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L29)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

It was actually intentional :) Wanted to focus reader's attention one detail at a time - it might be easier to understand. But can revert it if you think it's okay in one snippet.
</blockquote></details>

I think in this case is easier to grok the code with a well commented snippet.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156)*

juanjux:
Review status: 3 of 5 files reviewed at latest revision, 75 unresolved discussions.

---

*[driver/transform-dsl.md, line 136 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-136:-LAS9Bax9bRMnTGSR9Y0:bhzgyst) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L136)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

No, it still requires a `Map`. Should I write the full snippet with `Map` to make it more clear?
</blockquote></details>

Yes, I think it would be clearer that you need to do it inside a Map in that way.

---

*[driver/transform-dsl.md, line 374 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-374:-LAS9UBEBu7Lm9-9ZsXR:b8lgpgw) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L374)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

> Also, Opt can be used not only for nil values but also for values that can be missing, right?

If by "missing values" you mean "missing fields", then no. There was a note above:

> Note that the DSL makes a distinction between null fields and fields that does not exist.

Now I added few more details. Is it more clear this way?
</blockquote></details>

Yes, I think it's fine now.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156)*

juanjux:
Random idea: (this could be in a separate PR) it would be great to make a "recipes" page with some snippets like the ones you showed me these days to solve specific issues (doing different annotations depending on the value of a string, creating a new subnode to contain a list, etc.)

juanjux:
Reviewed 1 of 5 files at r1, 1 of 2 files at r2.
Review status: 4 of 5 files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156)*

dennwc:
Was actually planning to add it to this PR with some examples of role annotations.

---

Reviewed 2 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-:-LAYeF5cAYNDEMCWUMKE:bivaz4q)*

juanjux:
So, outside the snippets and other improvements, this is in pretty good shape but I guess we should wait for the SDK to be merged before merging this.

---

Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-:-LAlUpiC-jMKiW1IJT0W:bu5am1p)*

driver/transform-dsl.md
juanjux: What do you think about this?

dennwc:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[driver/transform-dsl.md, line 549 at r1](https://beta.reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-549:-LAljeMkDW2cxB7rtCST:b-lrinbk) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L549)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

What do you think about this?
</blockquote></details>

Will add it a bit later - it's indeed a very important topic.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/documentation/156)*

juanjux:
Reviewed 8 of 8 files at r4.
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/156)*

dennwc:
Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[driver/annotations.md, line 163 at r4](https://reviewable.io/reviews/bblfsh/documentation/156#-LCEVldvD0IjKXf4vU38:-LCcTRVx9Wd9Za0ay8jX:bqj541m) ([raw file](https://github.com/bblfsh/documentation/blob/5d7e853672caa277e12709ea841f0f6ea04e338b/driver/annotations.md#L163)):*
> I'm not sure if the last part is true for this helper anymore
No, this helper is unordered, but it's OK.

---

*[driver/transform-dsl.md, line 478 at r1](https://reviewable.io/reviews/bblfsh/documentation/156#-LANO1mjl6BQ0BrSn9rN-r1-478:-LCcf58LCxKPk8546kqs:b-896fix) ([raw file](https://github.com/bblfsh/documentation/blob/8ad32e023b8943832c6c505803b2c170fb936eae/driver/transform-dsl.md#L478)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Since Field is pretty important, could use one of those expanded examples with the json representations of the objects before and after the transformation.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/156)*

dennwc:
@juanjux I would say we should merge it as-is and add other docs as a separate PRs.

juanjux:
There is a link that is broken (check CI), fix that and I'll happily merge.

PR #160, opened 2018-06-18
Show the need to check Response.Status

juanjux:
From comments on: bblfsh/sdk/issues/257#issuecomment-398075978

cc @smola 


dennwc:
Still something wrong with the links

PR #168, opened 2018-07-09
Improve BIP3 rendering

ceh:
Signed-off-by: Emil Hessman <emil@hessman.se>


ceh:
Sample rendering before:
screen shot 2018-07-09 at 20 52 58

Rendering after:
screen shot 2018-07-09 at 20 52 29

PR #183, opened 2018-08-10
Add info to bblfshctl parse

### LINKED ISSUE #179: Running Babelfish locally for user files
Related and previous conversation at https://github.com/src-d/devrel/issues/77

juanjux:
Fixes src-d/devrel/issues/77
Fixes #179 

cc @tsolakoua 

### DIFF
--- a/using-babelfish/getting-started.md
+++ b/using-babelfish/getting-started.md
@@ -96,11 +96,10 @@ To test the driver you can execute a parse request to the server with the `bblfs
 $ docker exec -it bblfshd bblfshctl parse /opt/bblfsh/etc/examples/python.py
 ```
 
-If you want to parse instead one file in your local filesystem, mount the
-directory containing the files on the docker image with the `-v` option or
-creating a volume but the easiest way to do it would be to use one of
-the [clients](using-babelfish/language-clients.md) that would also allow you to
-do XPath queries over the results.
+Use this only for testing the installation; if you want to do any real parsing
+with files in your local filesystem you should use one of the
+[clients](language-clients.md) which would also allow you to run XPath queries
+over the results.
 
 ### Running standalone
 
--- a/using-babelfish/getting-started.md
+++ b/using-babelfish/getting-started.md
@@ -96,6 +96,12 @@ To test the driver you can execute a parse request to the server with the `bblfs
 $ docker exec -it bblfshd bblfshctl parse /opt/bblfsh/etc/examples/python.py
 ```
 
+If you want to parse instead one file in your local filesystem, mount the
+directory containing the files on the docker image with the `-v` option or
+creating a volume but the easiest way to do it would be to use one of
+the [clients](using-babelfish/language-clients.md) that would also allow you to
+do XPath queries over the results.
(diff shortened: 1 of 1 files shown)


using-babelfish/getting-started.md
```diff
@@ -96,6 +96,12 @@ To test the driver you can execute a parse request to the server with the `bblfs
 $ docker exec -it bblfshd bblfshctl parse /opt/bblfsh/etc/examples/python.py
 ```
 
+If you want to parse instead one file in your local filesystem, mount the
+directory containing the files on the docker image with the `-v` option or
+creating a volume but the easiest way to do it would be to use one of
```
tsolakoua: - creating -> create. 
- suggestion: However, the easiest way to do it would be...

using-babelfish/getting-started.md
```diff
@@ -96,6 +96,12 @@ To test the driver you can execute a parse request to the server with the `bblfs
...
+the [clients](using-babelfish/language-clients.md) that would also allow you to
```
tsolakoua: page not found

using-babelfish/getting-started.md
```diff
@@ -96,6 +96,12 @@ To test the driver you can execute a parse request to the server with the `bblfs
...
+do XPath queries over the results.
```
tsolakoua: suggestion: do -> run

dennwc: Maybe also propose to download the client and `bblfsh-cli` to be available to send local files instead of dealing with Docker volumes, which might be hard to remember?

juanjux: That would be better, on one hand it's simpler and would not confuse users into thinking that's is the canonical way to parse source files, on the other hand this is the section of testing a driver install while parsing and querying is covered elsewhere, what do you think @tsolakoua ?

tsolakoua: 1. I'm not familiar with Docker, so I need to do some search on my own to understand how mount/volumes work. I don't know the level of an average Babelfish user, but we might need to take into account the fact that maybe some people, like students haven't played with Docker yet. In general, that part confused me a bit, and I don't know if it's because of lack of knowledge from my side or because it's confusing. 
2. I agree with @dennwc , maybe it could simplify parsing local files.
3. @juanjux parsing and querying indeed are another thing as I understand and we don't need to get into details there, but I think in the Getting Started section maybe it's good to have an easy way to make the user at least parse/run something.

juanjux: It won't be simpler from the point of view of installation instructions since it'll be a two step process (install the Go/Python/Scala client and parse the file) instead of one which also complicates the flow of the section. I think I'll change the content to:

- Remark more strongly that the command is just to test the installation with one of the examples inside the image.
- Removing any reference to mounting volumes.
- Remark that any real parsing should be done with a client.

juanjux:
Rephrased that part again, please re-check.

juanjux:
Thanks to you @tsolakoua for the diligent work you're doing with this.

PR #184, opened 2018-08-10
Running docker section reorganized

### LINKED ISSUE #181: Stateless vs Non-stateless execution
Previous conversation at https://github.com/src-d/devrel/issues/78#issuecomment-412038229.

Seems it's not very clear the separation of stateless and non-stateless (or stateful? ) in the [documentation](https://doc.bblf.sh/using-babelfish/getting-started.html). 

I hadn't understood that after the sentence "On macOS, you first need to create etc." is referring to non-stateless mode. In my opinion is a bit confusing the non-clear separation. Maybe a sentence could be added saying something like: 

- For non-stateless execution you can do the following:
  - For MacOS...
  - For Linux...

juanjux:
Fixes #181 

I also reorganized the next section.

cc @tsolakoua 

### DIFF
--- a/using-babelfish/getting-started.md
+++ b/using-babelfish/getting-started.md
@@ -16,15 +16,27 @@ After playing with the dashboard, you will probably want to get Babelfish runnin
 
 ### Running with Docker \(recommended\)
 
-The easiest way to run the _bblfshd_ daemon is using Docker. You can start it with the following command:
+The easiest way to run the _bblfshd_ daemon is using Docker. You can do it in
+stateless mode, meaning that all installed drivers will be wiped once you stop
+the container, or using a Docker volume to store part of the container internal 
+filesystem and thus add persistence.
+
+#### Stateless mode
+
+For stateless mode, run the following command:
 
 ```bash
 $ docker run -d --name bblfshd --privileged -p 9432:9432 bblfsh/bblfshd
 ```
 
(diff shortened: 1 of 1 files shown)


using-babelfish/getting-started.md
```diff
@@ -16,15 +16,27 @@ After playing with the dashboard, you will probably want to get Babelfish runnin
 
 ### Running with Docker \(recommended\)
 
-The easiest way to run the _bblfshd_ daemon is using Docker. You can start it with the following command:
+The easiest way to run the _bblfshd_ daemon is using Docker. You can do it in
+stateless mode, meaning that all installed drivers will be wiped once you stop
```
dennwc: > drivers will be wiped once you stop the container
Since the commands don't involve `--rm` it would happen only on explicit removal of `bblfshd` container. Not sure that this makes any difference for the end user, but maybe worth mentioning.

juanjux: Good point, fixed.

using-babelfish/getting-started.md
```diff
@@ -17,7 +17,7 @@ After playing with the dashboard, you will probably want to get Babelfish runnin
 ### Running with Docker \(recommended\)
 
 The easiest way to run the _bblfshd_ daemon is using Docker. You can do it in
-stateless mode, meaning that all installed drivers will be wiped once you stop
+stateless mode, meaning that all installed drivers will be wiped once you remove
```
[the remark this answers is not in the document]
juanjux: Fixed, thanks!

PR #188, opened 2018-08-13
source codes -> source code

### LINKED ISSUE #187: Text improvement on clients documentation
In the [clients](https://doc.bblf.sh/using-babelfish/language-clients.html) guide: 

 ``` The client API's differ to adapt to their language specific idioms, the following [codes] [shows] several simple examples with the Go, Python and Scala clients that [parsers] a file and [applies] a filter to return all the simple identifiers. ```

-  I don't know if it's correct in english using the plural ```codes```. Maybe it could be replaced with ```code snippets``` for example. 
- ```shows``` should be ```show``` also
-  ```parsers``` -> ```parse```
-  ``` applies ``` -> ```apply```

As I'm not native english speaker, every time I mention something non-trivial please double-check if possible in case I am wrong.

juanjux:
Fixes #187

cc @tsolakoua 

### DIFF
--- a/using-babelfish/language-clients.md
+++ b/using-babelfish/language-clients.md
@@ -17,8 +17,8 @@ There are clients for the following languages:
 ## Examples
 
 The client API's differ to adapt to their language specific idioms, the
-following codes shows several simple examples with the Go, Python and Scala
-clients that parsers a file and applies a filter to return all the simple
+following code snippets show several simple examples with the Go, Python and Scala
+clients that parser a file and apply a filter to return all the simple
 identifiers:
 
 ### Go example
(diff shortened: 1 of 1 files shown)


using-babelfish/language-clients.md
```diff
@@ -17,8 +17,8 @@ There are clients for the following languages:
 ## Examples
 
 The client API's differ to adapt to their language specific idioms, the
-following codes shows several simple examples with the Go, Python and Scala
-clients that parsers a file and applies a filter to return all the simple
+following code snippets show several simple examples with the Go, Python and Scala
+clients that parser a file and apply a filter to return all the simple
```
tsolakoua: parser (noun) -> parse (verb)

PR #213, opened 2018-10-09
Clarify relation about @pos, Positions and Position

juanjux:
@zurk please check the changes and tell me if the relation is clearer now.

### DIFF
--- a/uast/semantic-uast.md
+++ b/uast/semantic-uast.md
@@ -40,8 +40,9 @@ present on any object regardless of the type (schema).
 
 This UAST specification defines few more special fields:
 
-* `@pos` - stores the positional information related to this UAST node.
-  See `Positions` type for more details.
+* `@pos` - stores the positional information related to this UAST node with a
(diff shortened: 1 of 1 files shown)


uast/semantic-uast.md
```diff
@@ -86,7 +74,21 @@ to access source snippet related to the node.
 
 As an example of a custom positional information, a ternary operator
 `x ? y : z` node may store individual positions for `?` and `:` characters
-as a separate `then` and `else` fields in `Positions` node.
+as a separate `then` and `else` fields in `Positions` node. This field will
+always be in the parent node under the `@pos` property.
```
dennwc: I haven't expected this to be confusing, but from @zurk's feedback it looks like it is. Maybe it's worth to add an example with the serialized `Positions` node? @juanjux What do you think?

juanjux: I was going to add a commented example of semantic UAST to the docs, but in a separate PR.

zurk:
oh, now I get it. Yes, it is much better, thank you @juanjux.

PR #226, opened 2018-12-04
Replace mermaid diagrams by PNG images

### LINKED ISSUE #163: mermaid diagrams are not rendered
Example:

https://docs.sourced.tech/babelfish/uast/uast-specification

juanjux:
Fixes #163.


juanjux:
@bzz I would like to keep automatic commits to a minimum. The idea of keeping the mermaid files on version control is interesting, but the point of mermaid is to be inlined into markdown, as a separate format there are better options like graphviz. Anyway since these are already done I'll add the source to the repo in the next commit.

Could you explain what is the external service you're using and how?

juanjux:
@bzz mermaid files added to the repo.

graphs_source/import_diagram1.mermaid
```diff
@@ -0,0 +1,11 @@
+graph TD
```
dennwc: This literally looks like graphviz. Maybe we should switch to it, as @juanjux mentioned? We already have some internal tools to dump UASTs in graphviz format, so it sounds like a logical step.

smola: Yep. Mermaid was used initially because it was easier to integrate with JS server and client side as well as with the old gitbook plugins. Now that we generate them upfront, whatever you choose for sources is ok.

juanjux:
All: I'll merge this one since there are only these two diagrams and both are currently broken. GraphViz would be a better option but we can start migrating to it if we have more and more complex diagrams.

PR #230, opened 2019-02-06
Update languages list

dennwc:


dennwc:
Yeah, I should have specified it in the description. It's only `make languages`.

We should automate it at some point.

Issue #236, opened 2019-03-05
Clarify UASTv2 status


mcarmonaa:
Related to the empathy session.

From the source{d} intro documentation you can get directly to https://docs.sourced.tech/babelfish/uast/uast-specification. From this page is not clear what is the status of UASTv1/UASTv2. Maybe a link to https://docs.sourced.tech/babelfish/#current-status would be helpful or even make the content of this talk about v2 and left a link to the legacy v1 instead of the other way.

creachadair:
I believe this should be addressed by #240. If I have missed anything you wanted to have addressed, feel free to reopen it and let me know.

PR #240, opened 2019-03-14
Make UAST v2 the default specification in documentation links.

creachadair:
This is an attempt to clarify the status of UAST v2 in response to #236. This is intentionally not a very large change—there is a lot of other cleanup we could do here but I have not tried to fold it together too tightly.

For now:
* Rename the UAST v1 and v2 spec files to a common format.
* Update the summary to link to v2 as the default,
* Update v1 to more strongly indicate that v1 is being replaced and that v2 should be referred, and
* Update existing spec links to point to the v2 base.

Suggestions are welcome.


creachadair:
I'm going to go ahead and merge this; @juanjux if you have any concerns I'm happy to send a follow-up.

PR #249, opened 2019-04-10
languages: re-generate the list of drivers

bzz:
Result of the `make languages`

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -3,7 +3,7 @@
 		"Name": "Bash",
 		"Language": "bash",
 		"Status": "beta",
-		"SDKVersion": "2.16.0",
+		"SDKVersion": "2.16.4",
 		"Documentation": {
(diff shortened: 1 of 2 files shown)


languages.json
```diff
@@ -3,7 +3,7 @@
 		"Name": "Bash",
 		"Language": "bash",
 		"Status": "beta",
-		"SDKVersion": "2.16.0",
+		"SDKVersion": "2.16.4",
```
creachadair: Oops, I guess we forgot to do this the last couple times. 😢

dennwc: It's quite hard to update on each point release of SDK without automation, so we usually update only if new release improves the performance or fixes something for this driver.

creachadair: Agreed—I wasn't complaining, but it's a good point in favour of finding a way to automate the check, at least (as we've discussed previously).

bzz:
Yup. And thank you for the prompt reviews.

PR #251, opened 2019-05-02
Add a UAST decoding spec

dennwc:
Add a UAST decoding spec written based on the design doc and Go implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/251)

### DIFF
--- a/uast/uast-encoding-v2.md
+++ b/uast/uast-encoding-v2.md
@@ -1,0 +1,251 @@
+# UAST v2 binary encoding
+
+The UAST binary encoding uses [Protocol Buffers](https://developers.google.com/protocol-buffers/)
+to encode individual messages, but requires additional logic to reconstruct the UAST.
+
+The design document can be found [here](https://docs.google.com/document/d/1iDOuWLnTckBdSscUMWGl2hBqAAqRn8jdRabXZSdr2aw/edit?usp=sharing)
+and the reference implementation in Go can be found in the [Babelfish SDK](https://github.com/bblfsh/sdk/blob/b44cc02214bf0d9c55dc3b830f423a6df6709ca9/uast/nodes/nodesproto/nodesproto.go).
+
+On a high level, a binary UAST uses [UAST v2 data model](./representation-v2.md).
+It extends this model by adding an ephemeral node ID field used to link UAST nodes.
+
+The file consist of two sections:
+
+* A [header](#header), containing a file format marker ("magic"), format version and other options.
+* A flat [list of nodes](#nodes) linked by node IDs.
+
+The list contains zero or more graph nodes addressed by ID. Binary format version `1` must
+only store tree structures, but the format was made compatible with more general graph structures
+for future extensions.
+
+The ID of the root tree node is stored in the [Header](#header) message.
+
+## Encoding
+
+### Messages
+
+Each protocol buffer message in the binary format is length-prefixed, similar to `bytes` values used in protocol buffers fields.
+
+First, a message size is encoded as an [unsigned varint](https://developers.google.com/protocol-buffers/docs/encoding),
+followed by `N` bytes of the message itself.
+
+There are two messages used to encode the UAST:
+
+- [GraphHeader](https://github.com/bblfsh/sdk/blob/b44cc02214bf0d9c55dc3b830f423a6df6709ca9/uast/nodes/nodesproto/nodes.proto#L17)
+- [Node](https://github.com/bblfsh/sdk/blob/b44cc02214bf0d9c55dc3b830f423a6df6709ca9/uast/nodes/nodesproto/nodes.proto#L46)
+
+See the sections below for the usage of those messages.
+
+### Header
+
+The binary UAST file begins with a `\x00bgr` (`0x00 0x62 0x67 0x72`) "magic" that can be
+used to detect a format of the file.
+
+It is followed by a version number encoded as an unsigned 32 bit integer in little-endian byte order.
+The current format version is `1`.
+
+Next, a length-prefixed `GraphHeader` message follows:
+```proto
+message GraphHeader {
+    uint64 last_id  = 1;
+    uint64 root     = 2;
+    uint64 metadata = 3;
+}
+```
+
+`last_id` is an optional field that sets a maximal node ID reserved by the node ID allocator.
+If not set, `max(node.id)` is assumed (maximal node ID in the [nodes list](#nodes-list)).
+Implementation may reserve more IDs than required by the UAST nodes by setting `last_id > max(node.id)`.
+
+`root` is an ID of the tree root node. Multiple unnamed roots can be stored by referencing an
+[Array](#array) node by its ID, and multiple named roots can be stored by referencing an
+[Object](#object) node. If not set explicitly, implementations should search the graph for
+unreferenced nodes of type [Array](#array) or [Object](#object) and treat them as an array of roots.
+
+`metadata` is an optional ID of a file metadata root, similar to `root`. This tree may
+contain arbitrary tree structure that describes the UAST file. If not set, no metadata is
+assumed. Implementation may ignore this tree, except for a case when `root` is not explicitly.
+
+### Nodes list
+
+Following the [Header](#header) message is a list of zero or more length-prefixed
+[Node](https://github.com/bblfsh/sdk/blob/b44cc02214bf0d9c55dc3b830f423a6df6709ca9/uast/nodes/nodesproto/nodes.proto#L46)
+messages, consuming the remainder of the file. There is no explicit node count provided by the format.
+If a corrupt node is encountered (e.g., length prefix bigger than remaining data, or node message invalid),
+this must result in a decoding error.
+
+### Node
+
+Each Node is described by the following protocol buffer message:
+```proto
+message Node {
+    uint64 id = 1;
+    
+    oneof value {
+        string string = 2;
+        int64  int    = 3;
+        uint64 uint   = 4;
+        double float  = 5;
+        bool   bool   = 6;
+    }
+    
+    repeated uint64 keys = 7;
+    uint64 keys_from = 10;
+    
+    repeated uint64 values = 8;
+    
+    bool is_object = 9;
+    
+    uint64 values_offs = 11;
+}
+```
+
+The `id` field is unique, monotonically increasing positive value. ID value is used in most message
+fields to reference other nodes.
+
+If `id` field is omitted, the `prevNode.ID + 1` is assumed. The first node with no `id`
+is assumed to have an ID value of `1`.
+
+*Note:* Nil is a special node distinct from an empty object or array, used to distinguish a pointer to
+an absent node from an absent pointer to a node. They are represented by ID value `0`.
+
+All other nodes can be of one of 3 kinds: [Values](#values), [Arrays](#array) and [Objects](#object).
+
+#### Values
+
+Values nodes must not have any other fields except `id` (optional) and `value` "oneof" field:
+
+```proto
+message Node {
+    uint64 id = 1;
+    
+    oneof value {
+        string string = 2;
+        int64  int    = 3;
+        uint64 uint   = 4;
+        double float  = 5;
+        bool   bool   = 6;
+    }
+}
+```
+
+Each "oneof" value corresponds to a value type defined in the [UAST v2 representation](./representation-v2.md#values).
+
+Value nodes may be deduplicated when encoding the UAST, thus implementations should make a
+copy of the value when decoding.
+
+#### Array
+
+Array nodes must only have an `id` (optional), `values` and `values_offs` fields set.
+`is_object` field may also be set to `false` when encoding an empty array.
+
+```proto
+message Node {
+    uint64 id = 1;
+    
+    repeated uint64 values = 8;
+    uint64 values_offs = 11;
+    
+    bool is_object = 9;
+}
+```
+
+`values` field contains a list of node IDs of the children nodes in the array. The list order
+corresponds to an array element order. Nil elements are encoded as ID value `0`.
+
+`values_offs` is an optional node ID offset that should be added to all `values` elements
+before resolving that child ID. Used for compression.
+
+Empty array are encoded by setting `id` field (optional) and optionally setting `is_object`
+to `false`. Message with no fields set is a valid empty array. `values` may be also set to
+an empty array and/or the `values_offs` may be set to `0`.
+
+#### Object
+
+Object nodes are represented by two arrays: one for object keys (`keys` and `keys_from`)
+and the second one for values (`values` and `values_offs`).
+
+```proto
+message Node {
+    uint64 id = 1;
+    
+    repeated uint64 keys = 7;
+    uint64 keys_from = 10;
+    
+    repeated uint64 values = 8;
+    
+    bool is_object = 9;
+    
+    uint64 values_offs = 11;
+}
+```
+
+Empty objects must have a `is_object` fields set. Other fields may be either set to zero value,
+or must not be set at all.
+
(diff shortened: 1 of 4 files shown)


uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+The `id` field is unique, monotonically increasing value. ID value is used in most message
+fields to reference other nodes.
+
+If `id` field is omitted, the `prevNode.ID + 1` is assumed. The first node with no `id`
+is assumed to have an ID value of `1`.
+
+Nil nodes must not be represented by the Node message. Instead, they are represented by
```
creachadair: (For separate discussion) I wonder if we shouldn't reserve a chunk of the ID space for well-known objects like the empty node (is that what "nil" is meant to be?) Maybe start real nodes at 1000 and keep 1..999 for plumbing?

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+Value nodes may be deduplicated when encoding the UAST, thus implementations should make a
+copy of the value when decoding.
+
+#### Array
+
+Array nodes must only have an `id` (optional), `values` and `values_offs` fields set.
+`is_object` field may also be set to `false` when encoding an empty array.
+
+```proto
+message Node {
```
creachadair: (For future consideration): In retrospect, we probably should consider making the mutual exclusion explicit in a later revision:

```proto
message Node {
   uint64 id = 1;
   oneof kind {
      Value value = 2;
      Array array = 3;
      Object object = 4;
   }

   message Value {
      oneof kind {
         string string = 1;
         // …
      }
   }
   message Array {
      uint64 offset = 1;
      repeated uint64 values = 2;
   }
   message Object {
      uint64 offset = 1;
      oneof keys {
         Keys keys = 2;
         uint64 keys_from = 3;
      }
      repeated uint64 values = 4;
   }
   message Keys {
      repeated uint64 keys = 1;
   }
}
```

Even though this duplicates some of the apparent structure, as a design property it's nice to have those type-distinct rather than just usage-distinct. It also makes it harder for consumers to get the field population wrong. (It is, however, a bigger change—and I'm not advocating we block anything for it, just raising the topic since we're paying attention to the code)

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+structures.
+
+It must also keep a set of seen nodes (`seen`) when reconstructing the tree to prevent loops.
+It must also keep a flag for each entry to this set to track if a given node is a tree branch,
+as opposed to a direct acyclic graph (DAG) branch. Nodes in tree branches are referenced
+only once, as opposed to DAGs where branches may overlap.
+
+Decoding process is can be the following:
+
+- The decoder read each node message from the file.
```
creachadair:
```suggestion
- The decoder reads each node message from the file in sequence.
```

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+- It assigns an `id` automatically from `current_id` and then incrementing the `current_id`,
+  or sets `current_id` to `id + 1` if the `id` field was set in the message.
+- It must ensure that `id` values are monotonically increasing, including both auto-assigned
```
creachadair: What if they are not? Should the reader fixup and backpatch the id sequence or report an error?

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+  and IDs specified in the messages.
+- It must ensure that ID is unique by checking it against the node map.
+- If `value` "oneof" is set, assume the node is a primitive value, add it to a map and read
+  the next message.
+- If `keys`, `keys_from` and `values` are either not set or zero, check the `is_object` field
```
creachadair: I think we use proto3, in which case there is no distinction between "unset" and "zero" for non-message-type fields (a message field can be set-empty though).

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+extensions.
+
+The ID of the root tree node is stored in the [Header](#header) message. Optionally an ID
+of a second tree root can be provided to store file metadata. Implementations are free to ignore it.
+
+## Encoding
+
+### Messages
+
+Each protocol buffer message in the binary format is length-delimited, similar to `bytes` values used in protocol buffers fields.
```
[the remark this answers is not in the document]
dennwc: Assuming this is ok, according to the comment below. I intentionally wanted to distinguish between the "file" (in our format) and "message" (individual protobuf message).

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+
+* [Header](#header), containing "magic", format version and other options.
+* Flat [list of nodes](#nodes) linked by node IDs.
+
+The list contains zero or more graph nodes addressed by ID. Binary format version `1` must
+only store tree structures, but the format was made compatible with graph structure for future
+extensions.
+
+The ID of the root tree node is stored in the [Header](#header) message. Optionally an ID
+of a second tree root can be provided to store file metadata. Implementations are free to ignore it.
```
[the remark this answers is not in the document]
dennwc: This was intended for the future if we decide to return some additional info with the AST (e.g. detected language, file name, etc). Right now it's unused. Assuming graphs, we may even drop it in the next spec revision.

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+    uint64 root     = 2;
+    uint64 metadata = 3;
+}
+```
+
+`last_id` is an optional field that sets a maximal node ID reserved by the node ID allocator.
+If not set, `max(node.ID)` is assumed. Implementation may reserve more IDs than required by
+the UAST nodes by setting `last_id > max(node.ID)`.
+
+`root` is an ID of the tree root node. Multiple unnamed roots can be stored by referencing an
```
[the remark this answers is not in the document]
dennwc: Sure, we use Arrays in the AST in general. But if you mean the array-of-roots, then no, we always specify a single root node in Babelfish.

This was done for a few reasons. First, we wanted a future-proof format (graphs = multiple roots =  no explicit root). But at the same time, we had a requirement from management that simple implementation of the decoder should be allowed to support trees only.

Thus, the format itself allows both behaviors: Babelfish emits the root pointer for compatibility, but other tools might not. Because of this, the spec is a bit vague here. It expects super naive implementation to fail without an explicit root, but more advanced implementations (that still expect a tree) can instead calculate an "array of roots" and use it as an implicit root node.

I agree that having this ambiguity is not ideal and we can discuss it separately.

As a note, if we decide to drop it, the Babelfish decode was always prepared for this and even old versions will automatically detect the root.

uast/uast-encoding-v2.md
```diff
@@ -0,0 +1,249 @@
...
+[Array](#array) node by its ID, and multiple named roots can be stored by referencing an
+[Object](#object) node. If not set explicitly, implementations should search the graph for
+unreferenced nodes of type [Array](#array) or [Object](#object) and treat them as an array of roots.
+
+`metadata` is an optional ID of a file metadata root, similar to `root`. This tree may
```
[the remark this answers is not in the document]
dennwc: Yeah, we should specify this somewhere. Maybe having a separate `concepts.md` is a good idea.

Answering the question, the metadata tree can be structured arbitrary and depends on the tool that emits the file. For now, we don't use it at all.

uast/uast-encoding-v2.md
dennwc: First, as a clarification, the `nil` node is different from an empty object (`{}`) and an empty array (`[]`). Maybe this distinction is not important, but I wanted to make JSON->UAST->JSON transformation lossless. This helps make reversibility easier.

Answering the question, I considered to include the "static table" that reserves some ID space. But then decided not to do so, because the optimal table may be different for Native/Semantic ASTs.

For example, some drivers won't emit empty object/arrays at all. Since the encoded ID length varies depending on the value (`unsigned varint`), having lower IDs helps with the overall file size. Following the example, implementations that don't use empty arrays may benefit from lower IDs by reusing the reserved space.

At the same time, possible 3rd-party implementations might not care about the file size and may emit a table in any order they prefer (e.g. DFS).

So the table order depends on whatever the encoder cares about space efficiency or not. If it does, it may emit the best possible table for a specific AST, while still conforming with the spec. If it doesn't care, it may do whatever it wants with it.

The general intention behind this format was that it should be vendor-agnostic, so it's not declaring any hardcode any Babelfish-specific predicates (even the `@type`). It may be useful in the long term. If the format survives, of course :grinning:

uast/uast-encoding-v2.md
dennwc: This was considered and was included in the first implementation, but the protobuf overhead for storing nested fields is too much, assuming the number of nodes in the typical AST. Thus, it was decided to inline fields. This makes decoding a bit harder since it's not handled by Protobuf directly. But since we already have a custom header and the format, it doesn't really matter (in my opinion).

As a note, we were discussing this previously. Maybe worth raising this issue again. But I think it may wait until we have some collaboration with other companies going on.

uast/uast-encoding-v2.md
dennwc: True, but technically the encoder may still emit fields with zero values, although it's discouraged in proto3. Still, proto3 decoder will still be able to read it (or at least it's the case for the decoder in Go). I decided that it may be better to specify what exactly happens in this case (specifically: they must be handled the same way).

dennwc:
@creachadair Thank you for all the valuable comments, everything is addressed, rebased and the PR is ready for another round.

Note that https://github.com/bblfsh/documentation/issues/255 still needs to be resolved, but let's make it a separate PR.

CI may fail, but should be fixed by https://github.com/bblfsh/documentation/pull/254.

uast/uast-encoding-v2.md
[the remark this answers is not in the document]
creachadair: Makes sense. Perhaps for this summary, then, we could simplify this to just the essential part:

```
The ID of the root tree node is located in the [Header](#header) message.
```

It doesn't preclude adding it back later, but the reader doesn't really need to do anything about it since it's unused and optional.

uast/uast-encoding-v2.md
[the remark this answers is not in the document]
creachadair: Re: Arrays in general vs. in the root. Yes, sorry for the ambiguity, I meant "do we use it in this context?"

It seems to me like we can simplify the spec, retain future-proofing, and reduce ambiguity for the implementor, by mandating that the reader must always scan for roots. For example maybe:

1. Any node in the list that is not referenced by any (other) node in the list is considered a _root candidate_.
2. If the `root` field of the header gives the ID of a root candidate, that candidate is treated as the root and the other root candidates are ignored.
2. If the `root` field of the header is 0, [1]
3. Otherwise (the `root` field does not point to a root candidate) the reader must report an error.

[Note 1]: There are options here: My intuition would be to say this is an error for now, and later when we add more graph structure we can relax that. That way a reader that expects one tree will fail predictably when we add a forest of trees and/or more graph structure, but can grow to accommodate it cleanly. 

Another option is to say that if `Header.root == 0` and there is exactly 1 root candidate, that must be used as the root, and if there are 0 root candidates or > 1 root candidates, it is an error. This is probably a good compatibility choice.

In any case, what I'm hoping to get at here is a clear specification of what the reader can expect. Right now we kind of allow "anything goes", there might be multiple roots or might not, and there might be a pointer to one of them, or to an array of them, and there's a bunch of correctness checks in there (do all the elements of the array point to root candidates? But then they're referenced inside the nodes list and aren't really root candidates; except that maybe this array is special because it's pointed to by `Header.root`? And so on).

uast/uast-encoding-v2.md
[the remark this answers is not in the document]
creachadair: Makes sense. Perhaps we shouldn't bother to mention it here, since it doesn't affect decoding?

uast/uast-encoding-v2.md
creachadair: Re: Nil vs. empty. Thank you for the clarification, I didn't realize that distinction was important. The rationale makes sense. We might call it out:

> **Note:** Nil is a special node distinct from an empty object or array, used to distinguish a pointer to an absent node from an absent pointer to a node.

or something along those lines.

I can see the motivation for allowing variation—e.g., in the order of outputs. But I think we need to be prepared to put at least _some_ constraints on the emitter: If we don't even have basic structural constraints, it will be hard to use the graph downstream without baking in all the possible variations and their meaning. I do think it makes sense to avoid constructs that might be hard to implement, though.

As a rule of thumb, we should try not to constrain too much about what the driver chooses to put into the nodes, but we should be clear about the basic topological data. This is not something we have to resolve here and now, though.

uast/uast-encoding-v2.md
creachadair: Now that you mention it, I do recall we talked about this back in November.

I believe it will be worthwhile to distinguish between _interchange format_ and _storage format_. Right now we are using the same message type for both. But I'd argue that for an interchange format it is more important to have a reliable API surface (one that is hard to use incorrectly), than to squeeze bits. It would be easy to mechanically transform nested messages (with their extra message-id tags) into flat messages or even something more compact than that (we did this in Kythe, for example https://godoc.org/kythe.io/kythe/go/storage/entryset).

An interchange format has _many_ clients: For us, the drivers produce it, the daemon consumes it, the normalization DSL manipulates it, the protocol clients receive it in responses and convert it to native data structures or JSON. My experience has been that having _only_ textual comments to document which fields have to be set together, and which are mutually exclusive, leads to big maintenance headaches: Updating the message structure to add new features, for example, often requires coordinating atomic changes across clients who all assume a particular set of rules.

For single files, even large ones, the extra bits for size tags almost certainly do not matter for interchange: gRPC compresses protocol messages anyway, and this stuff is all highly compressible. Of course, when we want to store a lot of these, the extra bits _could_ matter (though we should measure it). But then we can define a more compact storage format, private to the archiver, that knows how to go back and forth efficiently between that and interchange.

As you say, though, this is not a blocker for your PR here—only a discussion topic for us to revisit.

uast/uast-encoding-v2.md
creachadair: It's physically possible, but the spec says "[i]f a scalar message field is set to its default, the value will not be serialized on the wire" and all the implementations appear to respect this. More importantly, the API does not give the user any way to tell whether a scalar field or slice was present in wire encoding or not—you get the default value either way. So of course it's fine to mention it, but not really necessary 🙂.

uast/uast-encoding-v2.md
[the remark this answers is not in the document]
dennwc: I like the way you described the scan for root, but I have a few notes:
* Assuming future compatibility with graphs, the `root` may not be in the list of root candidates (it may be referenced by another graph node).
* If `root` is zero (`nil` tree), there is still a case when `metadata` may not be zero. If we won't mention `metadata`, the implementation will incorrectly pick it as root (since it's the only root candidate).

dennwc:
This is getting quite long and there are still topics to discuss, so I'm splitting a trivial part into #260. Will rebase this one, once it's merged.

bzz:
#260 seems to be merged now.

And I wonder, what would be the best place to move the https://github.com/bblfsh/documentation/pull/251#discussion_r290876554 discussion - I feel that this is pretty important feedback and it would be sad to just drop that accidentally.

bzz:
>> FIXME(dennwc): explain the v2 protocol, list relevant proto files and UAST decoding spec
>
> @bzz @creachadair Do you think it makes sense to do it as a separate PR?

Both ways make sense to me, whichever you prefer.

(somehow this discussion does not show up in reviewable for me, sorry)

PR #253, opened 2019-05-21
UAST type coverage report

bzz:
Initial version of the report that only includes UAST types, found in all driver's fixtures.
Triggered by `make types`.

Algorithm: 
 - discover all drivers, available on-line
 - clone/pull then in local FS
 - get available UAST types from SDK
 - for every driver, go though all `/fixtures/*.sem.uast` and count all `uast:*`

TODOs
 - for every driver, go though `normalizer.go` and count all usages of `uast.*`
 - get the list of all `uast.*` with go AST from SDK dynamically 

Example [report](https://github.com/bblfsh/documentation/blob/b56e5f7c25240fa36d9dc016eefbc2c31af19939/uast/types.md):
Screen Shot 2019-05-22 at 12 53 00 AM

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/253)


bzz:
CI failure is un-related

```
- /site/using-babelfish/getting-started.html
  *  External link https://www.docker.com/community-edition failed: 403 No error
```

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+	fmt.Fprintf(os.Stderr, "discovering all available drivers\n")
+	langs, err := discovery.OfficialDrivers(context.TODO(), &discovery.Options{
+		NoStatic: true,
+	})
+	if err != nil {
+		return nil, err
+	}
+	drivers := make([]*driverStats, 0, len(langs))
+	for _, l := range langs {
+		drivers = append(drivers, &driverStats{
```
dennwc: You may also want to call `l.IsRecommended()` to filter ancient drivers like Closure.

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+		return err
+	}
+
+	var (
+		cr = make(chan int, 3)
+		wg sync.WaitGroup
+	)
+	for i := range drivers {
+		wg.Add(1)
+		go maybeCloneOrPullAsync(drivers[i], &wg, cr)
```
dennwc:
```golang
i := i // copy the value
go func(){
  defer wg.Done()
  maybeCloneOrPullAsync(drivers[i], cr)
}()
```
It's a bit less elegant, but it's considered a good practice to include all usages of `WaitGroup` in one place (if possible).

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+// pulls the latest master for each of them.
+func maybeCloneOrPullAll(drivers []*driverStats) error {
+	fmt.Fprintf(os.Stderr, "cloning %d drivers to %s\n", len(drivers), reposRootPath)
+	err := os.MkdirAll(reposRootPath, os.ModePerm)
+	if err != nil {
+		return err
+	}
+
+	var (
+		cr = make(chan int, 3)
```
dennwc:
```suggestion
		cr = make(chan int, 3) // fetches to run in parallel
```
Or maybe there is a better name that can be used here.

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+
+func maybeCloneOrPull(d *driverStats) error {
+	repoPath := path.Join(reposRootPath, d.path)
+	_, err := os.Stat(repoPath)
+	if err != nil && !os.IsNotExist(err) {
+		fmt.Fprintf(os.Stderr, "error: %s\n", err)
+		return err
+	}
+
+	if os.IsNotExist(err) {
```
dennwc:
```golang
if os.IsNotExist(err) {
  // ...
} else if err != nil {
  // ...
}
```

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+// isUsedIn checks if current UAST type is used in the given package.
+func (u *uastType) isUsedIn() {
+	// TODO
+}
+
+// findUastTypesInSdk finds all types from SDK.
+func findUastTypesInSdk() []uastType {
+	var out []uastType
+	// TODO: load package, iterate all structs
+	types := []interface{}{
```
dennwc: Can be a global.

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+
+// analyzeFixtures goes though all fixtures, assuming the driver is cloned.
+// It updates given driverStats with results.
+func analyzeFixtures(driver *driverStats) error {
+	fixDir := path.Join(reposRootPath, driver.path, fixtureDir)
+	fixtureFiles, err := lsDir(fixDir)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	if os.IsNotExist(err) {
```
dennwc: Same as above (place this check first in `else if`).

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+		driver.skip = true
+		return nil
+	}
+	driver.fixturesNum += len(fixtureFiles)
+
+	re := regexp.MustCompile("(uast:[^\"]*)")
```
dennwc: `regexp.MustCompile` is slow and should be a global.

dennwc:
```suggestion
	re := regexp.MustCompile("uast:[^\"]*")
```
since there is only one group, and you are not using `FindSubmatch`

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+		if !x.skip {
+			drs = append(drs, x)
+		}
+	}
+	fmt.Fprintf(os.Stderr, "only %d drivers has fixtures, out of %d\n", len(drs), allDrivers)
+
+	formatMarkdownTableHeader(drs)
+	for _, typee := range uastTypes {
+		// %25s produces nice ASCII result
+		fmt.Printf("|[%s](%s)|", typee.name, goDocUrl+typee.name[strings.IndexRune(typee.name, '.')+1:])
```
dennwc: I think it may be a good idea to also replace `.` with `:` in the table, because it's the type name that we expose to our users in UAST.

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+var (
+	reposRootPath = path.Join(".", "drivers")
+	pprof         = flag.Bool("pprof", false, "start pprof profiler http endpoing")
+	skipUpdate    = flag.Bool("skip", false, "skip git clone or pull")
+)
+
+func main() {
+	flag.Parse()
+
+	if *pprof {
```
dennwc: I imagine this was to analyze some concurrency issues but can be removed now, right?

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+	fixturesNum  int
+	fixturesUast map[string]int
+	codeUast     map[string]int
+}
+
+// listDrivers lists all available drivers.
+func listDrivers() ([]*driverStats, error) {
+	fmt.Fprintf(os.Stderr, "discovering all available drivers\n")
+	langs, err := discovery.OfficialDrivers(context.TODO(), &discovery.Options{
+		NoStatic: true,
```
dennwc: Also probably `NoMaintainers` and `NoBuildInfo` to not fetch `MAINTAINERS`, `go.mod`, etc.

`NamesOnly` may have been better, but it won't allow checking for `IsRecommended`.

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	if os.IsNotExist(err) {
+		driver.skip = true
+		return nil
+	}
+	driver.fixturesNum += len(fixtureFiles)
+
+	re := regexp.MustCompile("(uast:[^\"]*)")
+	for _, file := range fixtureFiles {
+		if strings.HasSuffix(file.Name(), fixtureExt) {
```
dennwc: Please try to "lower" the code on identation as much as possible:
```golang
if !strings.HasSuffix(file.Name(), fixtureExt) {
  continue
}
// ...
```

_tools/types/main.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+// findUastTypesInSdk finds all types from SDK.
+func findUastTypesInSdk() []uastType {
+	var out []uastType
+	// TODO: load package, iterate all structs
+	types := []interface{}{
+		uast.Position{},
+		uast.Positions{},
+		// only embeding GenNode?
```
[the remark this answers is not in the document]
bzz: Sorry! This is meant to be the question about criteria to select types from UAST package: would it be useful to limit report only to the things that embed `GenNode` or just all the types found in that package? 

That would be a useful for further automation of type selection process, in case we want to do some filtering.

_tools/types/main.go
bzz: Please, help me understand - do you mean that the report should only include recommended drivers?

I'm not sure how valuable would that be - e.g for now it includes all the drivers that have fixtures, thus highlighting the scope of "further work".

_tools/types/main.go
bzz: True, but it's intentionally hidden to abstract the way it's being constructed: it's static hard-code for now \w a TODO for building this list dynamically.

_tools/types/main.go
bzz: Not sure I understand, from what I can see, now we expose both to the user:
  - `uast.Identifer` in API [godoc](https://godoc.org/github.com/bblfsh/sdk/uast#Identifier)
  - `uast:Identifier` in text [doc](https://doc.bblf.sh/uast/semantic-uast.html#identifier)

Or do you mean that in the final table, instead of current `uast.*` it's better to use convention of `uast:*`?  

Right now it's kind of consistent in using `.` AND linking to GoDoc, so in case of switching to `:`, would you also suggest to point the link to https://doc.bblf.sh instead?

_tools/types/main.go
bzz: As this is off by default and has no extra maintenance cost - I would prefer to keep the tooling handy.

_tools/types/main.go
bzz: Nice, I must have totally missed this awesome [Options doc](https://godoc.org/github.com/bblfsh/sdk/driver/manifest/discovery#Options).

_tools/types/main.go
bzz: addressed in 3b969c0

_tools/types/main.go
bzz: addressed in 3b969c0

bzz:
@dennwc thank you for the prompt review - all feedback has been addressed, it's ready for another round + would appreciate some feedback on a couple of open questions/improvement suggestions above.

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+package main
+
+import (
+	"bufio"
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"net/http"
+	_ "net/http/pprof"
```
creachadair: Please consider grouping "blank" (side-effecting) imports separately from the others.

creachadair:
```suggestion
```

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+	"os/exec"
+	"path"
+	"reflect"
+	"regexp"
+	"strings"
+	"sync"
+
+	"github.com/bblfsh/sdk/driver/manifest/discovery"
+	"github.com/bblfsh/sdk/uast"
+)
```
creachadair:
```suggestion

		_ "net/http/pprof"
)
```

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+			fmt.Printf(" %d/%d |", dr.fixturesUast[typee.name], dr.codeUast[typee.name])
+		}
+		fmt.Println()
+	}
+}
+
+func formatMarkdownTableHeader(drivers []*driverStats) {
+	fmt.Printf("|%25s|", "")
+	for _, dr := range drivers {
+		// %5s produces nice ASCII result
```
creachadair: Is this comment still relevant? (I ask because there is no `%5s` in evidence)

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+For every UAST type in every driver, next 2 numbers are reported:
+ - _fixtures usage_  - number of times this type was used in driver _fixtures_ (_*.sem.uast_ files)
+ - _code usage_ - number of times this type was usind in the driver mapping DSL code (_normalizer.go_ file)
+
+The format is <_fixtures usage_>/<_code usage_>.
+
+`
+
+const footer = `
+**Don't see your favorite AST construct represented? [Help us!](join-the-community.md)**
```
creachadair: Nice. 👍

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+	for range drivers {
+		fmt.Printf(" :-- |")
+	}
+	fmt.Println()
+}
+
+const header = `<!-- Code generated by 'make types' DO NOT EDIT. -->
+# UAST Types
+
+For every UAST type in every driver, next 2 numbers are reported:
```
creachadair:
```suggestion
For every UAST type in every driver, the following two values are reported:
```

(Rationale: Many prose style manuals recommend spelling out "small" numbers longhand, and switching to digits for larger ones. Chicago, for example, recommends spelling out 0–9 or 0–100 depending on context. This is not crucial, but that's why I made that part of the recommendation).

uast/types.md
```diff
@@ -0,0 +1,31 @@
+<!-- Code generated by 'make types' DO NOT EDIT. -->
+# UAST Types
+
+For every UAST type in every driver, next 2 numbers are reported:
```
creachadair:
```suggestion
For every UAST type in every driver, following two values are reported:
```

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+type uastType struct {
+	name string
+}
+
+// isUsedIn checks if current UAST type is used in the given package.
+func (u *uastType) isUsedIn() {
+	// TODO
+}
+
+// findUastTypesInSdk finds all types from SDK.
```
creachadair: I suggest: `findUASTTypesInSDK` per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+func findUastTypesInSdk() []uastType {
+	var out []uastType
+	// TODO: load package, iterate all structs
```
creachadair: Depending on what you mean here, I'm not sure this is feasible. The type information linked into the binary does not support iterating package contents. We could parse the package from source, but then we would have to generate the code for this declaration, since we would not have the type mirrors for those types.

That is possible, perhaps as a `go generate` rule, but tricky to get right.

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+		if err := scanner.Err(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func lsDir(dir string) ([]os.FileInfo, error) {
+	fmt.Fprintf(os.Stderr, "reading %s/*%s files\n", dir, fixtureExt)
+	f, err := os.Open(dir)
```
creachadair: https://godoc.org/io/ioutil#ReadDir may help here.

_tools/types/main.go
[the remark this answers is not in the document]
dennwc: I see what you mean now. I think we should list all the types including positions. We could omit those because we can't possibly miss them in new drivers, but those are only two types anyway, so it makes sense to include them for completeness.

_tools/types/main.go
dennwc: Hmm, you are right, the `IsRecommended` check is a bit too strict (`>= beta`). But maybe we could limit drivers to ones `ForCurrentSDK`? Or `SDKMajor > 1`? I just want to filter away all the drivers that are not supported by us currently.

_tools/types/main.go
dennwc: Yes, I think tables should list `uast:Identifier`, but still point to GoDoc (until we have a separate page describing the schema in Babelfish docs). 

The point is that most Babelfish users won't use Go, thus they will only see those types in the context of UAST/XPath (`uast:*`). The links to GoDoc serve as a schema definition and can be in any language. It could be a Protobuf definition or GraphQL for example. We happen to use Go as our DSL.

_tools/types/main.go
dennwc: I also thought initially that we should parse the code to get this list, but then I realized we [register](https://github.com/bblfsh/sdk/blob/master/uast/uast.go#L28) all those types in the `uast` package.

So I can expose an additional method in SDK to list registered types in a specific namespace and this code will get the list directly.

_tools/types/main.go
```diff
@@ -0,0 +1,340 @@
...
+	for _, file := range fixtureFiles {
+		if !strings.HasSuffix(file.Name(), fixtureExt) {
+			continue
+		}
+		
+		fFile, err := os.Open(path.Join(fixDir, file.Name()))
+		if err != nil {
+			log.Fatal(err)
+		}
+		defer fFile.Close()
```
dennwc: Defer in a loop will happen when the function returns. So if there are more than 1000 fixtures, you will hit file descriptor limit because of this leak.

_tools/types/main.go
bzz: I'm aware about [ReadDir](https://golang.org/src/io/ioutil/ioutil.go#L96) but it sorts the output. In our case, we do not care about an order and thus avoid wasting some CPU cycles.

_tools/types/main.go
bzz: I think it is - this is meant to be a reminder that replacing formatting string in the next line with this will give a nice ASCII instead of Md.

_tools/types/main.go
bzz: addressed in e820f75b45db947c7febff5fb8738af30a08fef6

_tools/types/main.go
bzz: Both sounds good, updated comment for clarity.

_tools/types/main.go
bzz: addressed in e820f75b45db947c7febff5fb8738af30a08fef6

bzz:
All feedback addressed, @dennwc it's ready for another round.

_tools/types/main.go
dennwc: Still needs to be added.

dennwc:
Github (again) does an awesome job of randomly hiding all comments including unresolved ones.

The change looks good, except for the driver filter (`ForCurrentSDK` and friends).

_tools/types/main.go
bzz: I must have let it slip, sorry!

> filter away all the drivers that are not supported by us currently.

Keeping only https://doc.bblf.sh/languages.html#supported-languages makes sence

bzz:
Somehow these must have slipped my mind:
 - update formating https://github.com/bblfsh/documentation/pull/253#discussion_r286669264
 - filter out more drivers https://github.com/bblfsh/documentation/pull/253#discussion_r286915625

Will addd asap.

_tools/types/main.go
creachadair: Okay, but I don't think these directories are very big, and this doesn't seem like a performance critical code path. If it does matter, though, I'd be fine with adding a comment to mention why we are bypassing the simpler solution. 😄

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+)
+
+const (
+	fixtureDir = "fixtures"
+	fixtureExt = ".sem.uast"
+	goDocURL   = "https://godoc.org/github.com/bblfsh/sdk/uast#"
+)
+
+var (
+	reposRootPath = path.Join(".", "drivers")
```
creachadair: For most uses, we should prefer `filepath.Join` (and the [filepath](https://godoc.org/path/filepath) package in general) for paths that are to be used with a real filesystem.

The difference is that `path` handles slash-separated paths, whereas `filepath` handles system-specific path layout. Since we do most everything on Linux this won't really make a difference in the observed results, but it follows the intended semantics of the package.

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	uastNodeRe    = regexp.MustCompile("uast:[^\"]*")
+
+	pprof      = flag.Bool("pprof", false, "start pprof profiler http endpoing")
```
creachadair:
```suggestion
	pprof      = flag.Bool("pprof", false, "start pprof profiler http endpoint")
```

Consider allowing the user to set the address for this, or at least document where it will run so they can tell from the `-help` text:

```go
   pprofPort = flag.Int("pprof", 0, "If non-zero start a pprof profiler HTTP endpoint at this port")

if *pprofPort > 0 {
   pprofAddr := fmt.Sprintf("localhost:%d", *pprofPort)
   …
}
```

or words to that effect?
```

File: _tools/types/main.go [hunk 25]
```diff
@@ -0,0 +1,341 @@
...
+		pprofAddr := "localhost:6060"
+		fmt.Fprintf(os.Stderr, "running pprof on %s\n", pprofAddr)
+		go func() {
+			if err := http.ListenAndServe(pprofAddr, nil); err != nil {
+				log.Fatal("cannot start pprof: %v", err)
+			}
+		}()
+	}
+
+	if err := run(); err != nil {
```
creachadair: We don't seem to rely on any important deferrals here—can we inline `run` in `main`?

(I wrote [ctrl](https://godoc.org/bitbucket.org/creachadair/ctrl) to handle this case for programs that need to do cleanup on exit, but here we don't really seem to need that)

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+func run() error {
+	drivers, err := listDrivers()
+	if err != nil {
+		return fmt.Errorf("failed to list drivers: %s", err)
+	}
+
+	if !*skipUpdate {
+		err = maybeCloneOrPullAll(drivers)
```
creachadair:
```suggestion
		if err := maybeCloneOrPullAll(drivers); err != nil {
```

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+		if err != nil {
```
creachadair:
```suggestion
```

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	formatMarkdownTable(drivers, uastTypes)
+	return nil
+}
+
+type driverStats struct {
+	url          string
+	language     string
+	path         string
```
creachadair: Please document what this is the path for.

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	skip         bool
```
creachadair: Please document what "skip" means.

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	fixturesNum  int
+	fixturesUast map[string]int
```
creachadair: [`fixturesUAST`](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

Also, please document what the keys/values are. Maybe a different name would help? `uastFeatureCount` perhaps?

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	formatMarkdownTable(drivers, uastTypes)
+	return nil
+}
+
+type driverStats struct {
+	url          string
+	language     string
+	path         string
+	skip         bool
+	fixturesNum  int
```
creachadair: [optional] I suggest `numFixtures` or `fixtureCount` or similar.

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	fixturesUast map[string]int
+	codeUast     map[string]int
+}
+
+// listDrivers lists all available drivers.
+func listDrivers() ([]*driverStats, error) {
+	fmt.Fprintf(os.Stderr, "discovering all available drivers\n")
```
creachadair: Minor nit:

```suggestion
	fmt.Fprintf(os.Stderr, "Discovering all available drivers ...\n")
```

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+		})
+	}
+	fmt.Fprintf(os.Stderr, "%d drivers available on-line\n", len(langs))
+	return drivers, nil
+}
+
+// maybeCloneOrPullAll either clones repos to path in local FS or, if already preset,
+// pulls the latest master for each of them.
+func maybeCloneOrPullAll(drivers []*driverStats) error {
+	fmt.Fprintf(os.Stderr, "cloning %d drivers to %s\n", len(drivers), reposRootPath)
```
creachadair: Minor nit:

```suggestion
	fmt.Fprintf(os.Stderr, "Cloning %d drivers to %q\n", len(drivers), reposRootPath)
```

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+func maybeCloneOrPullThrottled(d *driverStats, concurent chan int) error {
+	concurent <- 1
+	defer func() {
+		<-concurent
+	}()
+
+	return maybeCloneOrPull(d)
+}
+
+func maybeCloneOrPull(d *driverStats) error {
```
creachadair: The error from this is discarded. I think we should either return and handle it, or log it and drop it from the signature.

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+			maybeCloneOrPullThrottled(driver, pf)
+		}(drivers[i])
+	}
+	wg.Wait()
+	return nil
+}
+
+// maybeCloneOrPullThrottled is concurent-friendly verison of maybeCloneOrPull.
+// It "throttles" git operarions, based on given chanel capacity.
+func maybeCloneOrPullThrottled(d *driverStats, concurent chan int) error {
```
creachadair: I recommend against plumbing the concurrency down through the helpers. It's helpful for the reader to see all the concurrency matters in one place:

```go
throttle := make(chan struct{}, 3)
var wg sync.WaitGroup
for _, d := range drivers {
  d := d
  wg.Add(1)
  go func() {
     throttle <- struct{}{}
     defer func() { <-throttle ; wg.Done() }()
     maybeCloneOrPull(d)
   }()
}
```

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	return nil
+}
+
+type uastType struct {
+	name string
+}
+
+// isUsedIn checks if current UAST type is used in the given package.
+func (u *uastType) isUsedIn() {
```
creachadair: Is this meant to report a `bool`?

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	cmd := exec.Command("git", "pull", "origin", "master")
+	cmd.Dir = repoPath
+	err = cmd.Run()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type uastType struct {
```
creachadair: Could this be more simply:

```go
type uastType string

func (u uastType) isUsedIn(pkg string) bool { return … }
```
?

It does not appear to require mutability, and it's fine to use a plain string as the structure.

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+	driver.fixturesNum += len(fixtureFiles)
+
+	for _, file := range fixtureFiles {
+		if !strings.HasSuffix(file.Name(), fixtureExt) {
+			continue
+		}
+
+		fFile, err := os.Open(path.Join(fixDir, file.Name()))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to read %s, skipping\n", path.Join(fixDir, file.Name()))
```
creachadair: `continue` here?

_tools/types/main.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+
+		scanner := bufio.NewScanner(fFile)
```
creachadair: It looks to me like we have to read the whole file anyway, and these files are relatively small (just human-written source). On that basis, I would advocate a simpler strategy:

```go
data, err := ioutil.ReadFile(filepath.Join(fixDir, file.Name())
if err != nil {
  fmt.Fprintf(os.Stderr, "Unable to read %q, skipping\n", file.Name())
  continue
}
for _, line := range strings.Split(string(data), "\n") {
   …
}
```

It does cause the whole file to be read in, but for small files it may actually do _fewer_ allocations (not that this is perf critical anyway, since it's an occasional generation step). And it's simpler, not having to worry about whether the filehandle has been closed in various error conditions.

_tools/types/main.go
creachadair: That's fair. It's only that the function is specific to Markdown, so I couldn't see any way for the user to use this for a plain-text output. 😄

_tools/types/main.go
dennwc: Those fixtures may be relatively large, actually (tens of MB). So it should be faster to scan.

_tools/types/main.go
dennwc: Inlining will require pasting `log.Fatal(err)` everywhere, which is okay, but `return`s might be nicer :)

_tools/types/main.go
bzz: it is, but as noted in TODO, this is left for further improvement

_tools/types/main.go
bzz: Although true, I would prefer keeping it this way for consistency with the other `./_tool/`s

_tools/types/main.go
bzz: We usually have a separate flag for overriding the port, but as it's a small script I figured we can get by with the default port and a simpler usage, as soon as port is explicitly printed for convenience in terminal.

As it is meant to be run by `go run ..` - I would argue that it's alright to expect anyone interested in changing the port to just edit the default in line 44.

But please, let me know if you think we should add a add port overriding option though.

_tools/types/main.go
bzz: The needed fields will be decided at the time of implementing the `analyzeCode`, will add a comment.

_tools/types/main.go
bzz: Although I see the point, after 74 comments in review I'm starting to have either a self-doubt or a strange feeling that if we would have more people on the team reviewing this code - we would have even more opinions about how to write it, and thus will continue rewriting this simple, non-performance critical, trow-away script 😆

_tools/types/main.go
bzz: Was considering this, but decided in favor of buffered io, as I did not measure the performance of either approache.

_tools/types/main.go
bzz: This script uses the convention of not capitalizing anything. Please let me know if you have a strong preferences agains and then I'll change it everywhere for consistency.

Otherwise going to keep it as is.

_tools/types/main.go
bzz: added comment to go doc

bzz:
All feedback has been addressed, here is an example of the output:

Screen Shot 2019-05-23 at 8 03 27 PM

Script uses go modules now, in order to be runnable without GOPATH and avoid confusion with SDK versions.

_tools/types/main.go
creachadair: > Inlining will require pasting `log.Fatal(err)` everywhere, which is okay, but `return`s might be nicer :)

That's true—though the lint rule about error strings being uncapitalized and unpunctuated makes it ugly logs. Since we're already logging explicitly in a bunch of places, I thought it might be not so bad. To be clear, though, I am not extremely concerned about this, I just brought it up because it stuck out.

_tools/types/main.go
creachadair: Good point about the size. That said: I doubt it would be any faster—we're reading the whole file either way—the main savings is RAM. I don't think we really care about RAM usage here, it's a command-line tool run once in a while to update the stats, and we're only processing one by one.

_tools/types/main.go
creachadair: What you say makes sense. I agree we probably don't need to change the flag. But could you please include the port in the help text, so someone doesn't have to run the tool "for real" just to find out what the port is? E.g., `"Start a pprof profiler HTTP service at port 6060"` or words to that effect?

_tools/types/main.go
creachadair: Understood. But I feel like if we're going to check in the code, we should at least make sure it has the type signature we want, even if the implementation is just `return false`.

_tools/types/main.go
creachadair: > Although true, this seems to be a convention that all other `./_tool/`s use, and thus I would prefer keeping it this way for consistency.

Makes sense. I didn't realize we had done this in the others as well. (Maybe I'll import `ctrl` and clean these up later, but it's not important enough to make a mess for)

_tools/types/main.go
creachadair: It's purely an aesthetic recommendation, that's fine.

_tools/types/main.go
bzz: Added in 67e7dd961bccbd47efe815503e9b3882cc114c7c

_tools/types/main.go
bzz: Was changed to ioutil in 557d9b6

bzz: Thank you for sharing the assumptions and performance expectations though! It's such "layered" subject in modern systems that developing an intuition really takes a collective effort!

Did a few tests, on the current usage patten (scan by line + check & match)
  - `ioutil` version is consistent outperforming the `bufio` one ±2.20 sec
 - `bufio` version is ±2.40 sec

these are by no means an accurate benchmarks, just an observation on the time of processing all files, avg on 5 runs.

_tools/types/main.go
```diff
@@ -27,21 +27,23 @@ const (
...
+	pprofAddr  = "localhost:6060"
 )
 
 var (
 	reposRootPath = filepath.Join(".", "drivers")
 	uastNodeRe    = regexp.MustCompile("uast:[^\"]*")
 
-	pprof      = flag.Bool("pprof", false, "start pprof profiler http endpoing")
+	pprof = flag.Bool("pprof", false,
+		fmt.Sprintf("Start a pprof profiler HTTP service at %s", pprofAddr),
```
creachadair: Thanks!

Or simpler, if you like: `flag.Bool("pprof", false, "Start a pprof profiler HTTP service at "+pprofAddr)`

(this is fine too)

_tools/types/main.go
bzz: Results of removing variance due to network driver discovery and properly benchmarking `analyzeFixtures` on all 10 drivers:

```
Benchmark_bufio_ScanMatch-4   	   30000	     45404 ns/op	    1888 B/op	      50 allocs/op
Benchmark_ioutil_ScanMatch-4   	   30000	     41721 ns/op	    1888 B/op	      50 allocs/op
```

 - allocations are exactly the same
 - ioutil is a bit faster

🤔

_tools/types/main.go
bzz: There is a single right way in Go for everything, but a string interpolation 🤷‍♂ 
Thank you for simplification suggestion! History re-writtne in 58dc92b

_tools/types/main.go
creachadair: I don't know for sure, but I suspect the issue could be if your benchmark is using a file too small to really trigger any differences.

bzz: >  issue could be if your benchmark is using a file too small

Hm.. so far it uses the same data as for report generation: iterates over all 1001 `./fixtures/*.sem.uast` files in 10 different drivers. I'll dig deeper tomorrow and report back!

creachadair: The bytes-per-op are suspiciously small, and the allocations suspiciously uniform. I don't know what's going on there, but I think there must be something wrong in the way the benchmark is getting its data. (This is by no means crucial, so apart from curiosity please don't let this hold you up 😄).

bzz: Sorry for delay, here is an updated benchmark on single 500Mb UAST

```
Benchmark_bufio_ScanMatch_one-4		10	2390952772 ns/op	842338523 B/op	 9027097 allocs/op
Benchmark_ioutil_ScanMatch_one-4	10	1417128306 ns/op	1432769024 B/op	 4214468 allocs/op
```

`bufio` uses less memory but allocates more. `ioutil` is almost x2 faster but uses more memory.

PR #256, opened 2019-06-11
Update types report and link it from summary

bzz:
- re-generate type coverage report `make types`
 - link in from the summary page, so https://doc.bblf.sh/uast/types.html becomes a thing


bzz:
Thank you for prompt reviews! 
Merging after CI is happy.

Issue #255, opened 2019-06-05
Document DriverHost API (gRPC v2) 


dennwc:
We made a new Supported Languages API available in the gRPC protocol v2 and changed a few fields in the process. We should update docs to reflect this change.

bzz:
I suggest we also add a new column to the https://doc.bblf.sh/languages.html with the list of supported aliases - that would allow users to avoid possible confusion on figuring out which languages are supported by which drivers.

dennwc:
Mentioning aliases makes sense, although I'm wondering what is the use case for the end user? The driver either parses the file, or it's not. In the last case, we will get a new issue and will fix it by adding a new alias.

I don't think it's a bad idea, just want to save some space in the languages table :)

creachadair:
> I don't think it's a bad idea, just want to save some space in the languages table :)

Since I think not too many languages have aliases yet, we could maybe put a footnote in the table for each of the ones that do, listing the aliases. That way it'd still be close by to the table but not require a long column.

bzz:
> Mentioning aliases makes sense, although I'm wondering what is the use case for the end user?

the only use case is to inform the user, in order to prevent people from hard-coding the aliases and adding them on their side, assuming that otherwise `bblfshd` will not be able to parse it. Here is an [example](https://github.com/src-d/tm-experiments/blob/d0b3df6d171a2f5889e21c31d482d8993568a493/tmexp/utils.py#L15).

dennwc:
Oh, right, you mean to raise the awareness. OK, let me add the aliases then.

dennwc:
To give a bit more context here, the issue is about resolving the `FIXME` comment in [this file](https://github.com/bblfsh/documentation/blob/master/using-babelfish/babelfish-protocol.md).

We should properly document all  `DriverHost` methods and explain the difference between v1 and v2 methods (e.g. aliases, some fields missing, etc).

Later we will need to document the `Parse` API as well. Most of it is already done in https://github.com/bblfsh/documentation/pull/251, we only need to add a short description and reference that document. Probably deserves a separate issue.

kuba--:
Can we close it?

PR #264, opened 2019-07-04
Add a language aliases table

dennwc:
Add a (secondary) language aliases table, as discussed in #255.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/264)

### DIFF
--- a/_tools/languages/main.go
+++ b/_tools/languages/main.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"log"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/bblfsh/sdk/v3/driver/manifest"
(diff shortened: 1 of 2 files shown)


kuba--:
___
*[_tools/languages/main.go, line 133 at r1](https://reviewable.io/reviews/bblfsh/documentation/264#-LiyxvqQ0jsPOfy4Jiyn:-LiyxvqQ0jsPOfy4Jiyo:b-qf2uw3) ([raw file](https://github.com/bblfsh/documentation/blob/e491cf2e2dab291ee7ff4651038a724643e6c033/_tools/languages/main.go#L133)):*
> ```Go
> 		buf.WriteString("\n### Aliases for languages\n\n| Language | Aliases |\n| :--- | :--- |\n")
> 		for _, m := range list {
> 			if !m.ForCurrentSDK() || m.InDevelopment() || len(m.Aliases) == 0 {
> ```

maybe we can store these information in some slice from the first loop and simplify (get rid of the same condition check)  here - WDYT.

PR #274, opened 2019-07-31
Info about Supported languages v2

kuba--:
Signed-off-by: kuba-- <kuba@sourced.tech>
Related to: https://github.com/bblfsh/documentation/issues/255

This PR adds a short info about a new `SupportedLanguages` v2 requests.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/274)

### DIFF
--- a/using-babelfish/babelfish-protocol.md
+++ b/using-babelfish/babelfish-protocol.md
@@ -5,7 +5,31 @@ On the [next page](grpc-usage-example.md) we'll see a demo of how this comes tog
 
 ## Protocol v2
 
-<!-- FIXME(dennwc): explain the v2 protocol, list relevant proto files and UAST decoding spec -->
+Protocol v2 introduced a few changes in data structures and in methods' signatures.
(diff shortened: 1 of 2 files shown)


ncordon:
___
*[using-babelfish/grpc-usage-example.md, line 146 at r2](https://reviewable.io/reviews/bblfsh/documentation/274#-LlBR2kNBw-GzbE11R2B:-LlDPkDsDIm0i97N9FR-:b-6mn936) ([raw file](https://github.com/bblfsh/documentation/blob/144b5411315b2417e6d8f5df35bbb110ce3d9036/using-babelfish/grpc-usage-example.md#L146)):*
<details><summary><i>Previously, creachadair (M. J. Fromberger) wrote…</i></summary><blockquote>

It probably _does_ make sense to link to the data type in godoc.org (though are we still using gopkg.in for that package?). But I also agree that it's helpful to have an example: When I'm coding against an interface, the schema type is often not sufficient to really illustrate what I'm expecting, so examples are helpful.

I don't feel strongly that the example must be super-complete, having a rough outline for a simple but real query teaches me a lot.
</blockquote></details>

Personally I would also keep the example, as many people as @kuba-- said learn from examples.

PR #276, opened 2019-08-27
auto-update languages

### LINKED ISSUE #275: Implement driver's versions updater
We need a job that will update drivers versions in documentation once in a period(around a week).
Job should:
- get the list of versions of maintained drivers
- update the ones whos version was changed
- prepare related PR with changes.

Jenkins is more probably to be used.

lwsanty:
Once in a period jenkins job starts and performs `make languages command`
then it tries to create PR with language drivers versions update if needed.

Additionally curl trigger has been added just for the case when weekly cron trigger fails with rate limits or whatsoever

closes #275

Signed-off-by: lwsanty <lwsanty@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/276)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,58 @@
+pipeline {
+  agent {
+    kubernetes {
+      label 'bblfsh-docs-updater'
+      defaultContainer 'bblfsh-docs-updater'
+      yaml """
+spec:
+  nodeSelector:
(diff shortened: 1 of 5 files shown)


lwsanty:
@dennwc I guess I was checking that particular part isolated

PR #281, opened 2019-09-16
Automated update

jbeardly:
regular languages update

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/documentation/281)

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -73,14 +73,19 @@
 		],
 		"Maintainers": [
 			{
+				"Name": "Nacho Cordón",
+				"Email": "nacho@sourced.tech",
+				"Github": "ncordon"
+			},
+			{
+				"Name": "Kuba Podgórski",
+				"Email": "kuba@sourced.tech",
+				"Github": "kuba--"
+			},
+			{
 				"Name": "Denys Smirnov",
 				"Email": "denys@sourced.tech",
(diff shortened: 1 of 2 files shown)


creachadair:
The CI failed; I'm not sure if it's a transient failure, so I manually restarted it to see if it goes away on its own. If not, we might have a formatting issue (since it failed in the HTML checker).

creachadair:
It looks like the site cert for www.graphviz.org expired:

> This server could not prove that it is www.graphviz.org; its security certificate expired yesterday. This may be caused by a misconfiguration or an attacker intercepting your connection. Your computer's clock is currently set to Monday, September 16, 2019. Does that look right? If not, you should correct your system's clock and then refresh this page.

(my system clock is fine; I'm guessing Curl gets the same error)

creachadair:
https://gitlab.com/graphviz/graphviz/issues/1595

creachadair:
The site cert is still broken (~5h later). It is probably safe for us to merge this PR anyway, since the check that's failing is on the address rather than the structure of the doc. I'll let it sit till tomorrow morning and if it hasn't been fixed by then, I'll merge anyway.

creachadair:
https://gitlab.com/graphviz/graphviz/issues/1595 is resolved; I've manually re-triggered Travis.

PR #288, opened 2019-10-25
WIP Polishes docs

ncordon:
This is a general update to docs:

- [ ] Removes the clis for `python-client` and `scala-client` from the docs. We removed them from the corresponding packages because they supported different sets of options, Scala one had been broken for several versions, and we agreed the only cli to be used within a terminal should be the Go one.
- [ ] Review getting started section to update `bblfshd` to newly added rootless containers.
- [ ] Fixes link to join community (broken).
- [ ] Links template for design doc (the process for proposing new features was outdated with BIPs and we did not described a way of pitching in with design tasks). I made that link public in our GDrive, since it does not contain sensible information and it is just a template.
- [ ] Writes a CONTRIBUTE section with frequently asked questions: 
   - [ ] If I develop my own driver, how do I test it with `bblfshd`?
   - [ ] If I want to fix a failure in `libuast` that propagates to `scala-client` and `python-client`, how do I do it?  
   - [ ] How do I fix a failure in `sdk` and update it locally to test a `*-driver`? 
   - [ ] How do I modify and test `bblfshd` locally?
- [ ] Renames Babelfish web client to Babelfish playground in the index, since that terminology has been used before in the docs, and web client can be misleading with respect to `bblfsh-clients` for new users.
- [ ] Updates `tree-sitter` description, which does not reflect the actual state of that project.
- [ ] Updates the table of `clients` to reflect current state of things.
- [ ] Maybe dump in the docs the content of the post in our [forum](https://forum.sourced.tech/t/difference-between-uast-modes-native-annotated-semantic/69) about differences between ASTs, UASTs, and annotated ASTs.


ncordon:
Note: CI integration failing because we have to wait for https://github.com/bblfsh/bblfshd/pull/318 to be merged

